PR #232, opened 2023-10-05
Fix schema to allow descriptions-less apps

Malyuk-A:
Resolves this issue: https://github.com/oakestra/dashboard/issues/37

Demo: Seems to be working:
https://github.com/oakestra/oakestra/assets/65814168/3ed81f59-657e-43d6-a878-e0d59ae02681

Regarding other occurrences of "application_desc"
image

The only IMO interesting other place is here:

```python
def mongo_update_application(app_id, userid, data):
    db.app.logger.info("MONGODB - update data...")
    db.mongo_applications.find_one_and_update({'_id': ObjectId(app_id), 'userId': userid},
                                              {'$set': {'application_name': data.get('application_name'),
                                                        'application_namespace': data.get('application_namespace'),
                                                        'application_desc': data.get('application_desc'),
                                                        'microservices': data.get('microservices')}})
```

From my understanding `data.get('application_desc')` should not lead to issues because the get() method is a dictionary method that returns the default value (usually None) if the key is not found.

### DIFF
--- a/root_orchestrator/system-manager-python/sla/schema.py
+++ b/root_orchestrator/system-manager-python/sla/schema.py
@@ -122,7 +122,7 @@ sla_schema = {
                         "exclusiveMinimum": 0,
                     },
                 },
-                "required": ["applicationID", "application_name", "application_desc", "microservices"],
+                "required": ["applicationID", "application_name", "microservices"],
             }
         },
         "args": {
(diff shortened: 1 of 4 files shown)


root_orchestrator/system-manager-python/sla/schema.py
giobart, line 1: Inside the test folder, there are a couple of tests concerning the SLA, do they still pass? Is it possible to create a new one including an SLA without a service description?

giobart (changes requested):
Regarding the instance of `data.get('application_desc')`, what do you think if we change them to `data.get('application_desc',"")`, this way, if no application_desc we return empty string, just in case? Just to avoid getting the orchestrator to fail because of a None around the code :D
Let me know.

root_orchestrator/system-manager-python/sla/schema.py
Malyuk-A: Nicely spotted!
I have added the default value to the data.get and added a new SLA test (and refactored the test a bit)

giobart (approved):

giobart:
Fixes #37

Issue #61, opened 2022-04-11
Shrink the image sizes


giobart:
Let's see if we can shrink the release images. There are some dependencies and libraries that can be removed and take a lot of space.

Malyuk-A:
I had a look and wanted to share and discuss my findings:

The largest (custom) image currently is: system_manager with 1.05 GB

```
alex@alex-oakestra:~/oakestra$ docker images | grep system_manager
system_manager                                       latest    cd41e0709d06   2 minutes ago   1.05GB
```

It's Dockerfile uses this Baseimage: `FROM python:3.10`
This base image itself is already 1GB large:

`python                                               3.10      be529ebc3048   6 weeks ago     1GB`

So I tried a few other images.

Note: Via gitblame I found out that previously the same image was used here as it is still used for all other custom images, (python:3.8.5-slim-buster)

![image](https://github.com/oakestra/oakestra/assets/65814168/11bfda1a-c994-4e9e-b7a3-d58c75e5745d)
![image](https://github.com/oakestra/oakestra/assets/65814168/3298d3eb-9a74-47a0-a0e1-e052b9fe4f54)

This slim version is a lot smaller: (Because it expludes e.g. many of the UI-related Java libraries)
`python                                                  3.8.5-slim-buster   6cf621cb1327   3 years ago      113MB`

Maybe there was no slim version for the 3.10 out yet?

I tried it out with python:3.10-slim

```
python                                                  3.10-slim     83627dc9ab3c   6 weeks ago      128MB
system_manager                                          latest        038594b0f61c   2 minutes ago    174MB
```

I also tested python:3.10-alpine

```
python                                                  3.10-alpine   e8dedef264e4   6 weeks ago      49.7MB
system_manager                                          latest        c6cea54ce609   2 minutes ago    99.1MB
```

Also by simply modifying the pip install command a bit:

```
# Change: RUN pip install -r requirements.txt
# To:     RUN pip install -r requirements.txt --no-cache-dir && rm /requirements.txt
```

We get this:

`system_manager                                          latest        40538b980071   12 seconds ago       90MB`

I tested all of these variations by re-launching/building the different components (root-o., cluster-o., node (net & manager), logging in, creating a new application, new services and deploying them. Sofar it seems to work out.
I would appreciate if a maintainer could verify this with e.g. a more complicated test run.

If my assumption holds up, we could reduce our largest image (system_manager) from 1.05 GB to 90MB -> less than 10% of the original size!

[...]

PR #245, opened 2023-10-13
Shrink system-manager image size

Malyuk-A:
See this ticket for a detailed explanation: https://github.com/oakestra/oakestra/issues/61#issuecomment-1761519336

Let's try it out with alpine & busybox.
If it turns out to be too uncomfortable to debug we can simply switch to "slim" instead.


root_orchestrator/system-manager-python/Dockerfile
```diff
@@ -1,9 +1,9 @@
-FROM python:3.10
+FROM python:3.10-alpine
```
giobart, line 1: @Malyuk-A will replace this with -slim

Issue #169, opened 2023-03-07
Resource Abstractor Service


giobart:
## Short
A new service that abstracts generic information about a cluster or worker. 

## Proposal
Currently the schedulers directly use the cluster/worker database to retrieve the active clsuters (root schedulers) or active workers (cluster sschedulers)
We need a new microservice that interacts with the cluster/worker database and abstracts the available information. The service exposes an API that can be used to query the currently available resources. 

![schedulers-integration](https://user-images.githubusercontent.com/25887329/223424200-9de859e2-e595-4bd3-af2e-391218b9a4ad.png)

The schedulers will use this API to retrieve the available active worker nodes or clusters, as shown in the picture. 

Ideally, both root and cluster orchestrators will expose the same API, representing the aggregated cluster resources or a specific worker node as a generic resource. 
The GET method of the REST API returns the following:

- cpu cores (cluster & worker)
- memory (cluster & worker)
- gpu drivers (cluster & worker)
- location (cluster & worker)
- disk space (cluster & worker) 
- architecture list (cluster & worker)
- worker numbers (cluster only)

The service must envision:
- API versioning (as this service needs to be extended in the future with new hardware abstraction capabilities) 
- Custom return options, e.g., at root-level we return information such as worker-numbers, which is not necessary at cluster-level. In the future, we might have more such options.  
- OpenAPI specification automatically generated and exposed using swagger. 

## Ratio
This component will enable the possibility of integrating generic schedulers, as they only need to comply with the API. Solving this issue removes a dependency between the schedulers and the cluster/worker DB. 

## Impact
A new component at Root orchestrator and Cluster orchestrator. This can also be part of a separate repository. Discuss this option in the comments.

## Development time
1-2 weeks. 

## Status
- Looking for a candidate API and a solution draft.

## Checklist
 
- [x] Discussed
- [x] Documented
- [x] Implemented
- [x] Tested

PR #255, opened 2023-12-04
Resource Abstractor Service

melkodary:
#169 
**PR tackles**:
- [x] Creation of new abstract service.
- [x] Cloud Scheduler in Root utilizes the new service. Cloud Scheduler no longer access the database.
- [x] Add readme
- [x] Add new service to docker-compose.

**Futuer PR**:
- Next PR should include having the `system_manager` also utilizing the new service.
- Move jobs endpoint to another service.


root_orchestrator/resource-abstractor/Dockerfile
```diff
@@ -0,0 +1,21 @@
...
+ADD . /
+
+# TRUE for verbose logging
+ENV FLASK_ENV=development
+ENV FLASK_DEBUG=TRUE
+
+ENV CLOUD_MONGO_URL=localhost
+ENV CLOUD_MONGO_PORT=10007
+
+ENV MY_PORT=10011
```
Malyuk-A, line 17: Nitpick:
What does "MY" refer to here?
If it refers to the resource-abstractor (i.e. what is being built in this Dockerfile) wouldn't it be a bit clearer to use `RESOURCE_ABSTRACTOR_PORT` instead?

root_orchestrator/cloud-scheduler/cloud_scheduler.py
```diff
@@ -60,9 +58,9 @@ def replicate():
     job_id = data.get('job')
```
giobart, line 58: I think this endpoint /api/calculate/replicate is deprecated. Can you remove it?

root_orchestrator/cloud-scheduler/resource_management/cluster_operations.py
```diff
@@ -0,0 +1,39 @@
...
+
+def get_resources(**kwargs):
+    request_address = RESOURCE_ABSTRACTOR_ADDR + '/api/v1/resources'
+    try:
+        response = get(request_address, params=kwargs)
+        return response.json()
+    except exceptions.RequestException as e:
+        print('Calling Resource Abstractor /api/v1/resources not successful.')
+    
+    return None
```
giobart, line 14: I fear that this None can potentially result in some fatal error. After logging the error in case of exception, we can return an empty list of resources, I believe. This will then result in failed scheduling results to the orchestrator. We can then decide over there if and when to command a new reschedule.

giobart: Same thing for get_resource_by_x

root_orchestrator/cloud-scheduler/resource_management/job_operations.py
```diff
@@ -0,0 +1,36 @@
...
+
+def get_jobs(**kwargs):
+    request_address = JOBS_API
+    try:
+        response = get(request_address, params=kwargs)
+        return response.json()
+    except exceptions.RequestException as e:
+        print('Calling Resource Abstractor /api/v1/resources not successful.')
+    
+    return None
```
giobart, line 15: Same considerations as cluster_operations regarding the None

root_orchestrator/resource-abstractor/Dockerfile
giobart: I guess he's following the same structure as the other components. But as a general rule, I agree with you, we can use `RESOURCE_ABSTRACTOR_PORT` here and slowly change the other instances.

root_orchestrator/resource-abstractor/app.py
giobart, line 1: Can we rename this file?

giobart: Also, can you add a README.md with the OpenApi specs for these api v1?

giobart (changes requested):

root_orchestrator/resource-abstractor/app.py
```diff
@@ -0,0 +1,41 @@
...
+    },
+)
+app.register_blueprint(swaggerui_blueprint)
+
+for blp in blueprints:
+    api.register_blueprint(blp)
+
+@app.route('/', methods=['GET'])
+def hello_world():
+    return "Hello, World! This is the Resource Abstractor.\n"
```
Malyuk-A, line 38: I assume this is WIP? Maybe add a TODO here or remove it?

giobart: I would keep an empty debug endpoint like '/'. If we insert a health check, this can be used to test the response 200 from the service. What do you think?

root_orchestrator/resource-abstractor/db/mongodb_client.py
```diff
@@ -0,0 +1,24 @@
+import os
+from flask_pymongo import PyMongo
+
+MONGO_URL = os.environ.get('CLOUD_MONGO_URL')
+MONGO_PORT = os.environ.get('CLOUD_MONGO_PORT')
+
+MONGO_ADDR_CLUSTERS = 'mongodb://' + str(MONGO_URL) + ':' + str(MONGO_PORT) + '/clusters'
+MONGO_ADDR_JOBS = 'mongodb://' + str(MONGO_URL) + ':' + str(MONGO_PORT) + '/jobs'
```
Malyuk-A, line 8: I recommend using python's "f-strings" wherever possible. They are simpler and faster.

```suggestion
MONGO_ADDR_JOBS = f"mongodb:{MONGO_URL}:{MONGO_PORT}/jobs"
```

For more info see: https://www.geeksforgeeks.org/formatted-string-literals-f-strings-python/

root_orchestrator/resource-abstractor/start-up.sh
```diff
@@ -0,0 +1,19 @@
+#!/bin/bash
+
+# docker run -p 6379:6379 -d redis
```
Malyuk-A, line 3: Is this WIP or can this be removed?

root_orchestrator/resource-abstractor/start-up.sh
```diff
@@ -0,0 +1,19 @@
...
+
+# create virtualenv
+virtualenv --clear -p python3.8 .venv
+source .venv/bin/activate
+
+.venv/bin/pip install -r requirements.txt
+
+export FLASK_ENV=development
+export FLASK_DEBUG=False # TRUE for logging
```
Malyuk-A, line 12: Nitpick: Is this comment necessary? 😅 _(I get hammered at work regularly for my comments that are rather self-evident, so I start seeing this myself)_

root_orchestrator/cloud-scheduler/resource_management/resource_helper.py
```diff
@@ -0,0 +1,12 @@
+from _datetime import datetime
+
+CLUSTERS_FRESHNESS_INTERVAL = 30
+
+def is_cluster_active(cluster):
+    print('check cluster activity...')
+    timestamp_now = datetime.now().timestamp()
+    last_modified_cluster = cluster.get('last_modified_timestamp')
+    if last_modified_cluster >= timestamp_now - CLUSTERS_FRESHNESS_INTERVAL:
```
Malyuk-A, line 9:
```suggestion
    return last_modified_cluster >= timestamp_now - CLUSTERS_FRESHNESS_INTERVAL
```

root_orchestrator/cloud-scheduler/resource_management/resource_helper.py
```diff
@@ -0,0 +1,12 @@
+from _datetime import datetime
+
+CLUSTERS_FRESHNESS_INTERVAL = 30
+
+def is_cluster_active(cluster):
+    print('check cluster activity...')
```
Malyuk-A, line 6: Because we are already using loggers in many places I think it would also make sense to use them here:
Nice resources: 
- https://docs.python.org/3/howto/logging.html
- https://docs.python.org/3/library/logging.html

@giobart do we have any "inofficial" guidelines regarding logging? E.g. when/is it ok to use plain prints?

root_orchestrator/resource-abstractor/app.py
Malyuk-A: That makes sense, I would prefer to rename/refactor it then so it is obvious that it is a health check and not a potentially forgotten hello_world. 😅

root_orchestrator/cloud-scheduler/resource_management/cluster_operations.py
melkodary: Good call!

root_orchestrator/resource-abstractor/app.py
melkodary: My reason for `app.py` because it signifies that this is the entrypoint, so how about `resource_abstractor.py`?

root_orchestrator/resource-abstractor/app.py
melkodary: Agreed, `health` check point makes more sense.

root_orchestrator/resource-abstractor/db/mongodb_client.py
melkodary: Agreed! I think I was following the pattern used in the repo, given also that this url is being used in another module.

root_orchestrator/resource-abstractor/start-up.sh
melkodary: Good catch! I'll remove it.

root_orchestrator/resource-abstractor/start-up.sh
melkodary: Totally agree with you there. I think there are a lot of repetitive files, ie start-up.sh & Dockerfile, and this is just a case of copy & paste.

root_orchestrator/cloud-scheduler/resource_management/resource_helper.py
melkodary: This was copied from some other file, so didn't really look at the implementation of the file. However, I agree with you that we should have a more elagant way of logging. Happy to use the logger if we have a particular way of logging in the project.

root_orchestrator/cloud-scheduler/resource_management/cluster_operations.py
melkodary: But for single resource/item, shouldn't we return None if request fails? Also for the case `get_resource_by_id` the caller is checking if the result is None, so I suppose it should be fine, no?

root_orchestrator/cloud-scheduler/Dockerfile
```diff
@@ -19,6 +19,9 @@ ENV SYSTEM_MANAGER_PORT=10000
 ENV CLOUD_MONGO_URL=131.159.24.210
 ENV CLOUD_MONGO_PORT=10007
 
+ENV RESOURCE_ABSTRACTOR_URL=131.159.24.210
```
melkodary, line 22: Just a guess here. Would the upcoming `abstractor-service` container run there? @giobart

root_orchestrator/cloud-scheduler/resource_management/resource_helper.py
```diff
@@ -0,0 +1,8 @@
+from _datetime import datetime
+
+CLUSTERS_FRESHNESS_INTERVAL = 30
```
giobart, line 3: Ideally if the resource abstractor is the one responsible for the resources, it should be his job to tell us if a cluster is active or not. In my opinion, we cannot ask each scheduler to have its own cluster-freshness logic.

giobart (changes requested):

giobart (approved):

giobart:
@melkodary, all good; just make sure to resolve these conflicts, and we can merge this 💫💫

PR #259, opened 2023-12-08
Fix system-manager-python Docker image build

Malyuk-A:
Fixes the following issue:
image
image
image
image
Log: https://github.com/oakestra/oakestra/actions/runs/7085706626/job/19411364416rl

Resolves this: https://github.com/oakestra/oakestra/issues/61#issuecomment-1845368687

The issue was that the greenlet pip package requires gcc & g++ to work properly which were missing in the 3.10-slim or even 3.10-slim-buster base image but present in the large/previous python:3.10 image.

I have temporarily added a github workflow to build the image but not push it to verify that building the image is working on the github runner.


giobart:
Fixes #61

giobart (approved):

PR #266, opened 2023-12-18
migration to CRI interface

### LINKED ISSUE #264: migration from containerd based implementation to CRI
## Short
Right now, our code directly uses `containerd` sdk. It's better if we transition to CRI APIs, and then we attach `containerd` as the default runtime. 

## Proposal

The Container Runtime Interface (CRI) defines the main [gRPC](https://grpc.io/) protocol for the communication between a node and a generic container runtime. All the container runtimes provide CRI compliant plugins. This means that if we transition from `containerd` specific APIs, to CRI, we can dynamically swap the container runtime as we like. 

This is an image from the official containerd CRI docs:
cri

The CRI APIs can be imported using `runtimeapi "k8s.io/cri-api/pkg/apis/runtime/v1"`.

Here we have the proto file for the gRPC calls of CRI v1 ->  [proto](https://github.com/kubernetes/cri-api/blob/master/pkg/apis/runtime/v1/api.proto)

While [this](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/cri/remote/remote_runtime.go) is the current runtime implementation of k8s, which we can take as a reference. 

## Ratio

- Extensibility/modularity -> No need to manually add support for additional container runtimes
- Easier to extract container metrics and statistics: right now, we manually interact with cgroups files. This complexity is abstracted by CRI. 
- Easier APIs -> higher level APIs to interact with containers. 

## Impact
NodeEngine/virtualization/ContainersManagement.go

## Development time
1 week

## Status
looking for feedback

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

giobart:
fixes #264


giobart:
Closing for inactivity

PR #269, opened 2024-01-18
268 missing resource abstractor from 1 doc

giobart:
#fix 268

### DIFF
--- a/run-a-cluster/1-DOC.yaml
+++ b/run-a-cluster/1-DOC.yaml
@@ -106,10 +106,13 @@ services:
       - SYSTEM_MANAGER_PORT=10000
       - CLOUD_MONGO_URL=localhost
       - CLOUD_MONGO_PORT=10007
+      - RESOURCE_ABSTRACTOR_URL=localhost
+      - RESOURCE_ABSTRACTOR_PORT=11011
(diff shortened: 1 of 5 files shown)


run-a-cluster/1-DOC.yaml
```diff
@@ -95,10 +106,13 @@ services:
       - SYSTEM_MANAGER_PORT=10000
       - CLOUD_MONGO_URL=localhost
       - CLOUD_MONGO_PORT=10007
+      - RESOURCE_ABSTRACTOR_URL=localhost
+      - RESOURCE_ABSTRACTOR_PORT=11011
       - REDIS_ADDR=redis://:cloudRedis@localhost:6379
     depends_on:
       - redis
       - mongo_root
```
melkodary, line 114: cloud_scheduler no longer depends on mongo_root.

melkodary: we could also remove the mongo env from cloud_scheduler

melkodary (approved):

run-a-cluster/1-DOC.yaml
giobart: ty for noticing, removed those ;)

PR #272, opened 2024-02-05
256 allow management traffic to use ipv6

giobart:
Fixed ipv6 family on server sockets


smnzlnsk (approved):
Looks good to me!

nitindermohan (approved):
Accepted. 

NM

PR #273, opened 2024-02-07
Ipv6 compatibility fix

smnzlnsk:
- Added IPv6 support for resource abstractor
- Fixed IPv4 request incompatibility issues with request.remote_addr, which would cause it to be a 4to6 mapped IPv6 address
- Added overrides for IPv6 in **non**-1-DOC deployments
- Added override for local service manager builds (useful in automation scenarios)


cluster_orchestrator/override-ipv6-enabled.yml
giobart, line 1: what's the difference between enabled and addressed?

cluster_orchestrator/override-local-service-manager.yml
giobart, line 1: Let's not push this without further instructions. It might just confuse people.

root_orchestrator/override-local-service-manager.yml
giobart, line 1: Let's not push this without further instructions. It might just confuse people.

root_orchestrator/system-manager-python/ext_requests/cluster_requests.py
```diff
@@ -30,7 +13,7 @@ def cluster_request_to_deploy(cluster_id, job_id, instance_number):
     try:
         cluster_addr = (
             "http://"
-            + add_brackets_if_ipv6(cluster.get("ip"))
+            + sanitize(cluster.get("ip"), request=True)
```
giobart, line 16: 👍👍

giobart (changes requested):

cluster_orchestrator/override-ipv6-enabled.yml
smnzlnsk: Is documented in the README
Basically it changes the hostname inter-container addressing to static IPv6 addressing.
This way one can assure that a IPv4-to-IPv6 mapped IP address cannot be the case from inter-container traffic.
We can also remove it, as we addressed and fixed this.

cluster_orchestrator/override-local-service-manager.yml
smnzlnsk: Added instructions to the README!

cluster_orchestrator/override-ipv6-enabled.yml
giobart: Ok then if it's not necessary let's remove it

smnzlnsk: I removed it, since it most likely is not needed.

giobart (approved):

smnzlnsk:
Merged with #280

Issue #276, opened 2024-02-14
Unikernel scheduling resource calculation is broken


smnzlnsk:
## Short
Unikernel applications are not possible to be deployed 
## Proposal 
`cloud_scheduler` `calculate` always returns `NoActiveClusterWithCapacity` since calculation data is faulty.
`aggregation = cluster.get("aggregation_per_architecture", None)` (https://github.com/oakestra/oakestra/blob/develop/root_orchestrator/cloud-scheduler/calculation.py#L135) always returns `None`, because `resource_abstractor` does not return a `aggregation_per_architecture` field. 

## Reproduction
Deploy a cluster with 1 root, 1 cluster and 1 worker (node net manager + node engine with -u flag enabled [requires qemu to be installed])
Deploy a unikernel application using the SLA:
```
{
	"sla_version": "v2.0",
	"customerID": "Admin",
	"applications": [{
		"applicationID": "",
		"application_name": "nginx",
		"application_namespace": "test",
		"application_desc": "Simple demo of an Nginx server Unikernel",
		"microservices": [{
			"microserviceID": "",
			"microservice_name": "nginx",
			"microservice_namespace": "test",
			"virtualization": "unikernel",
			"cmd": [],
			"memory": 100,
			"vcpus": 1,
			"vgpus": 0,
			"vtpus": 0,
			"bandwidth_in": 0,
			"bandwidth_out": 0,
			"storage": 0,
			"code": "https://github.com/Sabanic-P/app-nginx/releases/download/v1.0/kernel.tar.gz",
			"arch": ["amd64"],
			"state": "",
			"port": "80:80",
			"addresses": {
				"rr_ip": "10.30.30.26"
			},
			"added_files": []
		}]
	}]
}
```

cloud_scheduler reports `NoActiveClusterWithCapacity` because the `cluster` returned by the resource_abstractor does not contain the field `aggregation_per_architecture`, which is required for calculation.

## Solution 
Add required fields to database operation in resource_abstractor and return it to cloud_scheduler

## Status
searching for solution

## Checklist
- [x] Discussed
- [x] Solved
- [x] Tested

smnzlnsk:
@melkodary Maybe you can take a look into it. Cheers.

PR #277, opened 2024-02-17
Unikernel scheduling resource calculation is broken

melkodary:
#276 

- [x] Partially Tested: _Only resource abstractor was tested and not the entire flow._


smnzlnsk (approved):
Works in my testing!

melkodary:
@giobart Could you have a look?

giobart (approved):
Yes, this fixes the issue; I'm able to fully deploy unikernels, as you can see from the logs I get from the dashboard 🎉

image

Issue #278, opened 2024-02-21
Pre-Push hook for linting


melkodary:
## Short
Git pipelines often fail because of linting issues, which then requires another commit to fix it, leaving the history of commits just filled with "liniting fix".

## Proposal
One solution is to just squash to correct the history but then you'll need to force push which is not really ideal. The proposed solution is to have a pre-push hook configured. This way locally one knows that they can't push unless the linting issues are fixed.

## Ratio
Not  a priority, but fairly easy to win and its a quick win.

## Impact
No impact

## Development time
<1 hr

## Checklist
 
- [x] Discussed
- [x] Documented
- [x] Implemented

Malyuk-A:
Out of curiosity, why are git force pushes not ideal?  😅

> ... are but then you'll need to force push which is not really ideal.

E.g. this is the git workflow of the company I work for and we use force pushes.
[git-workflow.pdf](https://github.com/oakestra/oakestra/files/14358559/git-workflow.pdf)

melkodary:
> Out of curiosity, why are git force pushes not ideal? 😅
> 
> > ... are but then you'll need to force push which is not really ideal.
> 
> E.g. this is the git workflow of the company I work for and we use force pushes. [git-workflow.pdf](https://github.com/oakestra/oakestra/files/14358559/git-workflow.pdf)

Well you need to be careful to not override remote commits and one might forget to use `with lease` option. Point is one has to be very careful about force pushing. Also the idea of rewriting history is something I try to avoid. I understand that people sometimes they have to force push anyways because of rebasing but that's another discussion I guess of `rebasing` vs `merging`.

PR #279, opened 2024-02-21
Pre-Push hook for linting

melkodary:
### PR:
#278

Using external tool called [pre-commit](https://pre-commit.com/) _(it also supports other hooks)_ that automatically will create the hooks - `.git/hooks/pre-${stage}`. 

Since pre-git hooks are not synced with developers or remote repo, developers who would like to utilize it must install the package `pip install pre-commit` and then execute it once:
```
$ pre-commit install --hook-type pre-push
```

### Status
- [x] Tested
- [x] Documented in [Oakestra Documentation](https://github.com/oakestra/documentation/pull/12) repo


Malyuk-A:
LGTM 👍

So basically all we need to do to get this working on our ends is to run?

```
pip install pre-commit
pre-commit install --hook-type pre-push
```

Would be nice to add this to the dev docs so that we do not forget this and the onboarding of new people will get smoother. 🙂

melkodary:
> LGTM 👍
> 
> So basically all we need to do to get this working on our ends is to run?
> 
> ```
> pip install pre-commit
> pre-commit install --hook-type pre-push
> ```
> 
> Would be nice to add this to the dev docs so that we do not forget this and the onboarding of new people will get smoother. 🙂

Yes exactly! I will add them to documentation repo, or do you mean we should also add it somewhere else?

Malyuk-A:
> I will add them to documentation repo, or do you mean we should also add it somewhere else?

Just adding this to the documentation repo should be fine IMO.

giobart (approved):

Issue #281, opened 2024-02-25
Make system_manager use resource abstractor


melkodary:
## Short
In task #169, a new service resource abstractor was created and the `root_orchestrator::cloud_scheduler` utilized it. This task is just a continuation of the previous one.

## Proposal
The system manager should also utilize the resource abstractor to get access to resources.

## Ratio
This is important for further development for hooks & custom resources.

## Impact
Impacts the system manager and anything dependent on it - probably a breaking change.

## Development time
\>2 week

## Checklist
- [x] Discussed
- [x] Documented
- [x] Implemented
- [x] Tested

oakestra-admin:
🥀 This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

PR #282, opened 2024-02-25
Make system_manager use resource abstractor

melkodary:
#281 

- [x] Figure out how to handle making the new library available.
- [x] Add unit/integration test cases.


libraries/rasclient/__init__.py
Malyuk-A, line 1: What exactly does "ras" in rasclient stand for? I assume that it does not mean "Remote Access Service" but rather Resource Abstractor Service - client? If the latter is the case could we maybe call it "resource_abstractor_client" so that it is easier /faster to understand?

libraries/rasclient/client_helper.py
```diff
@@ -0,0 +1,20 @@
...
+
+
+def make_request(method, api, **kwargs):
+    url = f"{RESOURCE_ABSTRACTOR_ADDR}{api}"
+    try:
+        response = method(url, **kwargs)
+        response.raise_for_status()
+        return response.json()
+    except exceptions.RequestException:
+        print(f"Calling {url} not successful.")
```
Malyuk-A, line 18: Do we already have some kind of logger (that we could use here) that we could use here or should we really continue using prints for this? If I am not mistaken Luca mentioned that he will introduce something that will have the same "interface" as the python logging util, so maybe we could already use it here to avoid future heavier refactoring.

root_orchestrator/resource-abstractor/db/apps_db.py
```diff
@@ -0,0 +1,38 @@
...
+
+def update_app(app_id, data):
+    return db.mongo_apps.find_one_and_update(
+        {"_id": ObjectId(app_id)},
+        {"$set": data},
+        return_document=True,
+    )
+
+
+def _create_app(app_data):
```
Malyuk-A, line 29: Why is this separate helper function needed? Correct me if I am wrong, but isn't it only used in the method below it so couldn't we simply fuse them?

root_orchestrator/resource-abstractor/db/clusters_db.py
```diff
@@ -1,7 +1,24 @@
+from datetime import datetime
+
 import db.mongodb_client as db
 from bson.objectid import ObjectId
+from db.clusters_helper import get_freshness_threshhold
```
Malyuk-A, line 5: Nitpick: typo
-> threshold

root_orchestrator/resource-abstractor/db/clusters_db.py
```diff
@@ -13,14 +30,58 @@ def find_clusters(filter):
...
+    datetime_now_timestamp = datetime.timestamp(datetime_now)
+
+    cpu_percent = data.get("cpu_percent")
+    cpu_cores = data.get("cpu_cores")
+    memory_percent = data.get("memory_percent")
+    memory_in_mb = data.get("cumulative_memory_in_mb")
+    nodes = data.get("number_of_nodes")
+    gpu_cores = data.get("gpu_cores")
+    gpu_percent = data.get("gpu_percent")
+    # technology = data.get('technology')
```
Malyuk-A, line 54: Nitpick: Could you add a comment to this commented in line explaining why it is commented it?
Can be helpful when you or someone else returns to this in the future, otherwise, this will become guesswork.

root_orchestrator/resource-abstractor/db/jobs_db.py
```diff
@@ -1,14 +1,86 @@
...
+                "instance_list.$.status_detail": job_data.get(
+                    "status_detail", "No extra information"
+                ),
+                "instance_list.$.logs": job_data.get("logs", ""),
+            },
+        },
+    )
+
+
+def _create_job(job_data):
```
Malyuk-A, line 65: Same here. What is the reason for isolating this? - just curious

root_orchestrator/resource-abstractor/db/jobs_db.py
```diff
@@ -1,14 +1,86 @@
...
+def create_update_job(job_data):
+    job_name = job_data.get("job_name")
+    job = db.mongo_jobs.find_one({"job_name": job_name})
+
+    if job:
+        return update_job(str(job.get("_id")), job_data)
+    else:
+        return create_job(job_data)
+
+    return job
```
Malyuk-A, line 86: If I am not mistaken this line will never be reached? - i.e. can be removed

root_orchestrator/resource-abstractor/tests/api_test.py
```diff
@@ -0,0 +1,36 @@
...
+from flask import Flask
+
+
+class BlueprintTestCase(unittest.TestCase):
+    def setUp(self):
+        self.app = Flask(__name__)
+        self.app.register_blueprint(applicationsblp)
+        self.client = self.app.test_client()
+
+    @patch("api.v1.apps_blueprint.apps_db.find_apps")
```
Malyuk-A, line 16: Question regarding how this works (note: I am not new to writing tests but it has been a while):
Do I understand this correctly?:
- You want to check the "find_apps" method int he "apps_blueprints.apps.db" API.
- You mock this method - thus replacing it entirely - via the mock_find_apps which simply returns the mock return_value

If so how exactly does this test the API endpoint if it gets substituted by the mock entirely? Or does it only substitute the underlying DB interacting but the find_apps method is called thus properly (unit) tested?

root_orchestrator/system-manager-python/services/instance_management.py
```diff
@@ -1,28 +1,49 @@
...
+        for instance in instances:
+            job_operations.update_job_instance(job_id, instance["instance_number"], instance)
+
+    return job_operations.update_job_status(job_id, status, status_detail)
+
+
+def update_job_status_and_instances(
+    job_id, status, next_instance_progressive_number, instance_list
+):
+    print("Updating Job Status and assigning a cluster for this job...")
```
Malyuk-A, line 28: Same here - plain print

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -44,16 +65,24 @@ def create_services_of_app(username, sla, force=False):
     return {"job_id": str(last_service_id)}, 200
 
 
+# TODO: have logging moved to resource abstractor
+def delete_job(job_id):
+    logging.log(logging.INFO, "MONGODB - delete job...")
+    job_operations.delete_job(job_id)
+    logging.log(logging.INFO, "MONGODB - job {} deleted")
+    # return mongo_frontend_jobs.find()
```
Malyuk-A, line 73: Can this line be removed?

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -44,16 +65,24 @@ def create_services_of_app(username, sla, force=False):
     return {"job_id": str(last_service_id)}, 200
 
 
+# TODO: have logging moved to resource abstractor
```
Malyuk-A, line 68: Why was the logging not yet moved?

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -44,16 +65,24 @@ def create_services_of_app(username, sla, force=False):
...
 def delete_service(username, serviceid):
-    apps = mongo_get_applications_of_user(username)
+    apps = app_operations.get_user_apps(username)
     for application in apps:
         if serviceid in application["microservices"]:
             # undeploy instances
             request_scale_down_instance(serviceid, username)
             # remove service from app's services list
             remove_service_from_app(application["applicationID"], serviceid, username)
             # remove service from DB
```
Malyuk-A, line 84: Nitpick: I know that this is legacy code and not your new one, however it would be nice to remove trivial/self-explanatory comments like these ones in this method.

root_orchestrator/system-manager-python/system_manager.py
```diff
@@ -104,11 +104,13 @@ def handle_init_client(message):
 
     cluster_address = sanitize(request.remote_addr)
 
-    cid = mongo_upsert_cluster(cluster_ip=cluster_address, message=message)
-    x = {"id": str(cid)}
+    message["cluster_ip"] = cluster_address
+    cluster = cluster_operations.create_cluster(message)
+    cid = str(cluster["_id"])
+    x = {"id": cid}
```
Malyuk-A, line 110: Could you maybe place this "x" content directly into the " emit("sc2", json.dumps(x), namespace="/register")" below?
Otherwise, this is just some random/cryptic var.
i.e. ~  emit("sc2", json.dumps({"id": cid}), namespace="/register")

root_orchestrator/system-manager-python/tests/docker-compose.yml
Malyuk-A, line 1: What exactly is the purpose of this compose? If it is just for testing, why can't the regular one be used for it?

root_orchestrator/system-manager-python/tests/services_test.py
```diff
@@ -1,278 +1,268 @@
...
-def test_create_service_with_app():
-    # SETUP
-    mockdb()
 
-    sla = get_full_random_sla_app()
-    db_app_mock = copy.deepcopy(sla)
+@pytest.fixture(scope="session")
+def resource_abstractor():
+    with DockerCompose(current_dir) as compose:
+        service_uri = "http://localhost:21011"  # TODO: have it fetched from compose object.
```
Malyuk-A, line 32: Could you do this already?

root_orchestrator/system-manager-python/tests/services_test.py
```diff
@@ -1,278 +1,268 @@
...
+        last_service = microservices[len(microservices) - 1]
+        last_service = generate_db_structure(created_app, last_service)
+
+        # EXEC
+        result, code = create_services_of_app("Admin", sla)
+
+        # ASSERT
+        assert code == 200
+        db_app_result = app_operations.get_app_by_name_and_namespace(
+            sla["applications"][0]["application_name"],
```
Malyuk-A, line 59: Nitpick/question-of-preferences:
In general this is written explicitly very often:  sla["applications"][0]
It is fine and all. How about creating a tiny helper function that gets the first app and calling it and when we have this  sla["applications"][0] multiple times in the same context (e.g. same function) fetch the application once and reuse it instead of calling  sla["applications"][0] over and over?

This is up to you if you want to do this or not. (It just feels a bit off to me)

root_orchestrator/system-manager-python/tests/docker-compose.yml
Malyuk-A, line 1: Also how hard do you think is it for you to add an automatic CI trigger for running these tests when related file changes occur. This would be great.

root_orchestrator/system-manager-python/tests/services_test.py
Malyuk-A, line 1: Because we are both using VSCode we have quite similar setups.
How do you run those tests?
I get this currently:
image

libraries/rasclient/__init__.py
melkodary: I think at first I had it as you suggested, i.e `resource_abstractor_client`, but I felt the name is just too long. Still you bring a good point that it may not be clear - I'll change it back then.

libraries/rasclient/client_helper.py
melkodary: I'll check out how luca is doing it. Perhaps we can tag him here to give us quick suggestions. Do you know his github handle, I am unable to tag/find him?

root_orchestrator/resource-abstractor/db/apps_db.py
melkodary: You'll totally right! There was a purpose where `create` actually would have a flag called `upsert` & `_create` is only strict to only creating, but I didn't proceed with that and I forgot to clean it up.

root_orchestrator/resource-abstractor/db/clusters_db.py
melkodary: Thanks! I actually always thought its `threshhold` similar to `withhold`. English is stupid 😄

root_orchestrator/resource-abstractor/db/jobs_db.py
melkodary: Same answer to the previous one.

root_orchestrator/resource-abstractor/tests/api_test.py
melkodary: It is indeed a simple test. The mocked function is the db and not the endpoint entirely. So the endpoint is responsible in returning a json and converting the `_id` to a string from `ObjectId` so basically serialization is being tested here. Do you think it is still not really needed?

root_orchestrator/system-manager-python/services/instance_management.py
melkodary: Will check with Luca.

root_orchestrator/system-manager-python/services/service_management.py
melkodary: Good question. Well I don't know yet how logging should be tackled. I feel its a discussion that we should have on what should be logged and where. Should every service log everything it's doing?

root_orchestrator/system-manager-python/services/service_management.py
melkodary: Definitely!

root_orchestrator/system-manager-python/services/service_management.py
melkodary: Sure thing!

root_orchestrator/system-manager-python/system_manager.py
melkodary: makes sense!

root_orchestrator/system-manager-python/tests/docker-compose.yml
melkodary: In my opinion, every environment should have it's own configuration, and this docker compose has a different configuration than the other one, for instance the ports exposed. If we use same configuration for dev & testing, the dev database will be polluted with testing records/documents. Also the test cases might fail because dev db might already be populated with data that may conflict with testing results.

root_orchestrator/system-manager-python/tests/services_test.py
melkodary: I believe I can. I'll fix it.

root_orchestrator/system-manager-python/tests/services_test.py
melkodary: Sure thing, it's more readable the way you suggested.

root_orchestrator/system-manager-python/tests/docker-compose.yml
melkodary: I didn't do it before, but I have a general idea of how to trigger a CI only for related file changes - basically you do a git diff between develop and your commit and checkout the files that have changed. I can see that it can be inaccurate sometimes and messy, so I prefer to always trigger all test cases regardless, unless it's a readme or a yaml file or something similar. I believe we already have the testing pipeline but I am not sure why it's not working.

root_orchestrator/system-manager-python/tests/services_test.py
melkodary: You need to install the `rasclient` library. Run:
```sh
pip install git+https://github.com/melkodary/oakestra.git@281-rasclient#subdirectory=libraries
```
_Note: I am still searching for an elagant way to do this._

libraries/rasclient/client_helper.py
Malyuk-A: I think he is not yet really part of the github project or so.
This is his account: https://github.com/TheDarkPyotr

root_orchestrator/resource-abstractor/db/clusters_db.py
Malyuk-A: Oh yes it is 😅

root_orchestrator/resource-abstractor/db/clusters_db.py
Malyuk-A: Ok than its fine for now, I guess.

root_orchestrator/resource-abstractor/tests/api_test.py
Malyuk-A: If it is not mocking the entire thing and the endpoint code is executed this seems fine to me.

root_orchestrator/system-manager-python/tests/docker-compose.yml
Malyuk-A: Ok makes sense, thanks.

Malyuk-A: Hmm ok, this can also be postponed to a separate PR to avoid bloating this one.

root_orchestrator/system-manager-python/tests/services_test.py
melkodary: Actually looking at it again, it seems the best thing i can do is do this:
```py
service_uri = (
            f"http://{compose.get_service_host('resource_abstractor', 21011)}"
            f":{compose.get_service_port('resource_abstractor', 20007)}"
        )
```
In which case I still need to specify the port. It is slightly better because it returns the mapped port - in case the mapping is not the same i.e 21011 doesn't map to 21011 and perhaps something else.

Malyuk-A: Hmm, I see. It's up to you I am fine with both, even though the simple localhost is more readable.

libraries/rasclient/client_helper.py
TheDarkPyotr: Not joined but reading here and there 😃 
For the logging approach at the moment the only approach I suggest is to use the standard `logging` library as the standardization process (*at the moment*) is a simple wrapper of the library but further discussion is needed. 
Next meeting surely will brought up the topic!

root_orchestrator/resource-abstractor/tests/api_test.py
Malyuk-A, line 1: image
I was able to test this successfully 👍

Malyuk-A (commented):
Changes seem reasonable to me.
Let's see what @giobart says.

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -1,38 +1,61 @@
...
         return {"message": "application not found"}, 404
     for microservice in data.get("applications")[0].get("microservices"):
         if not valid_service(microservice):
             return {"message": "invalid service name or namespace"}, 403
         # Insert job into database
         service = generate_db_structure(application, microservice)
-        last_service_id = mongo_insert_job(service)
+        last_service_id = insert_job(service)
+        if last_service_id is None:
+            # TODO: what should be done to previously inserted services?
```
giobart, line 53: Thanks for adding this check. 
I think previous services should stay where they are, nothing wrong with them. What should be done instead is:
- logging an error; this way, we trigger an alarm in the root dashboard. If there is an insert failure there is likely an issue in the cluster. 
- return in the response maybe a status code for each one of the services. like
```
{
  "job_id": str(last_service_id)  // leave it for compatibility with current dashboard
  "services":[
  {
       "service_id":id
       "status":int //status code 200/500/...
       "description":string //error description
      }
  ]
} 
```
The current dashboard anyway queries the system, but this might be helpful for come popup errors.

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -44,48 +67,58 @@ def create_services_of_app(username, sla, force=False):
...
-            mongo_delete_job(serviceid)
-            # inform network component
+            delete_job(serviceid)
             net_inform_service_undeploy(serviceid)
             return True
     return False
 
 
 def update_service(username, sla, serviceid):
     # TODO Check fields and redeploy service
```
giobart, line 88: Exactly! Can we temporarily return error with this method? I noticed that in the current version in develop and main it creates more issues than solutions. Until we have this check this looks to me as we're just creating issues.

giobart (changes requested):

root_orchestrator/system-manager-python/services/cluster_management.py
```diff
@@ -0,0 +1,21 @@
...
+
+    job_obj = job_operations.get_job_by_id(job_id)
+
+    if job_obj is None:
+        return None
+
+    instances = job_obj.get("instance_list")
+    if instances and len(instances) > 0:
+        instance = job_obj["instance_list"][0]
+        return cluster_operations.get_resource_by_id(instance["cluster_id"])
```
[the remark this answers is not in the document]
melkodary: You are right! I forgot to include the instance number in the request when fetching. Nice catch 🙌🏼

root_orchestrator/system-manager-python/services/instance_management.py
```diff
@@ -1,85 +1,125 @@
...
+    if job is None:
+        return None
+
+    for instance in instances:
+        job_operations.update_job_instance(job_id, instance["instance_number"], instance)
+
+    return job_operations.update_job_status(job_id, status, status_detail)
+
+
+def update_job_status_and_instances(
```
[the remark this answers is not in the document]
melkodary: Yeah it's confusing. Checkout files`apps_db` in develop the following functions `mongo_update_job_status_instances` & `mongo_update_job_status` - they are practically equivalent to `update_job_status` & `update_job_status_and_instances`. I am happy to change it to something more meaningful so let me know if you have suggestions 🙋🏼

root_orchestrator/system-manager-python/services/service_management.py
melkodary: Good Idea! I'll implement that.

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -1,91 +1,155 @@
...
+                }
+            )
+
         # TODO: check if service deployed already etc. force=True must force the insertion anyway
-    return {"job_id": str(last_service_id)}, 200
+    return {
+        "job_id": str(last_service_id),
+        "deployed_services": deployed_services,
+        "failed_services": failed_services,
+    }, 200
```
giobart, line 86: 👍

giobart:
@melkodary thank you for the PR! Code-wise green light on my side. I'll approve once tested 👍

root_orchestrator/system-manager-python/services/instance_management.py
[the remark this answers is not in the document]
giobart: Ok I found it and got it. Yes they can basically be merged but let's do it on a separate PR. Thank you!

giobart:
Hello, am I the only one experiencing the following issue when building the cloud scheduler from `root` docker compose? 

image

It basically refers to the git dependency form the cloud scheduler requirements.txt:

```
...
click~=8.1.7
redis~=5.0.1
requests~=2.27.1
werkzeug==2.0.3
git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries
```

melkodary:
> Hello, am I the only one experiencing the following issue when building the cloud scheduler from `root` docker compose?
> 
> image
> It basically refers to the git dependency form the cloud scheduler requirements.txt:
> 
> ```
> ...
> click~=8.1.7
> redis~=5.0.1
> requests~=2.27.1
> werkzeug==2.0.3
> git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries
> ```

Oh it seems I forgot to add git dependency to cloud_scheduler. I'll push a commit shortly. Please also note that this will still not work because this [PR](https://github.com/oakestra/oakestra/pull/292) needs to be merged first, since the package in requirement.txt is pointing to develop.

root_orchestrator/resource-abstractor/db/clusters_db.py
```diff
@@ -1,26 +1,88 @@
...
+    return db.mongo_clusters.find_one({"cluster_name": cluster_name})
+
+
+def update_cluster_information(cluster_id, data):
+    """Save aggregated Cluster Information"""
+
+    datetime_now = datetime.now()
+    datetime_now_timestamp = datetime.timestamp(datetime_now)
+
+    cpu_percent = data.get("cpu_percent")
```
giobart, line 49: Idea. Can we iterate through the keywords instead of manually setting them?, maybe define a translation map for the pre-defined known keywords such as cpu_cores. This way if a cluster was extended with additional capabilities such as... quantum cores, news sensors, or other stuff, we don't need to extend the abstractor. What do you think?

root_orchestrator/system-manager-python/services/service_management.py
```diff
@@ -27,68 +44,115 @@ def create_services_of_app(username, data, force=False):
...
+
+    #         job = job_operations.update_job(serviceid, sla)
+    #         if job is None:
+    #             logging.log(logging.ERROR, "job not updated")
+    #             continue
+
+    #         logging.log(logging.INFO, "job {} updated")
+    #         return job, 200
+
     return {"message": "service not found"}, 404
```
giobart, line 130: code 501 not implemented

root_orchestrator/system-manager-python/tests/docker-compose.yml
```diff
@@ -0,0 +1,28 @@
+version: "3.3"
```
giobart, line 1: why a compose file inside tests?

giobart (changes requested):

root_orchestrator/system-manager-python/tests/docker-compose.yml
melkodary: `services_test` uses it for integration testing - since it will spin up the containers as test containers along with a mongodb test container. Also the pipeline uses it. with module `testcontainers` they provide a way to spin up containers through docker-compose file which is nice.

root_orchestrator/resource-abstractor/db/clusters_db.py
melkodary: This could definitely be cleaned up. However we would still need to extend the translation map when new values are added, or perhaps I am missing something.

root_orchestrator/resource-abstractor/db/clusters_db.py
```diff
@@ -1,26 +1,95 @@
...
+        "gpu_cores": "total_gpu_cores",
+        "gpu_percent": "total_gpu_percent",
+        "virtualization": "virtualization",
+        "more": "more",
+        "worker_groups": "worker_groups",
+        "aggregation_per_architecture": "aggregation_per_architecture",
+    }
+
+    # Use the mapping to create the update dictionary
+    update_dict = {db_key: data.get(data_key) for data_key, db_key in key_mapping.items()}
```
giobart, line 64: doesn't it trigger an exception if one of the keys is missing from the update data?

melkodary: this is already the current implementation on `develop`: check out `cluster_db.py::mongo_update_cluster_information`. 
All I did was move this function somewhere else while renaming it and turning it into a map. Happy to update the logic if there is one.

giobart (approved):

PR #288, opened 2024-03-13
Add new one-shot service state

Malyuk-A:
# Abstract
Adds the new service state "COMPLETED" for services that are "one-shots", i.e. that are intended to run only once and not be persistent like e.g. a service. 

# Concrete Usecase
I am implementing an image builder that will be deployed on the workers. Once the image is built the app/service is no longer needed and can be undeployed. - This can be generalized, i.e. let's say a user wants to deploy a container that should only perform a specific action once and is no longer of any use afterwards, thus there is no need to redeploy it.

# Reasoning - more context
Turns out we currently only support persistent services and not one-shots.
Taken from the oakestra paper:
image
In fact the structure looks more like this:
image
A dead service currently gets redeployed all the time automatically by the cloud scheduler, until the service gets manually undeployed/removed.

Good news: The necessary SLA changes are not breaking but legacy compatible, because when we use a boolean with the default value of false for one_shot, the NodeEngine in go will simply handle it as false when it is undefined which is the case for all legacy/current services.

# Proposal
image
Add a new state for one-shot container use-cases. For now, it will simply have another state (COMPLETED) instead of DEAD thus the cloud scheduler will not try to resurrect it.

# Result
image


go_node_engine/model/Service.go
```diff
@@ -18,6 +18,7 @@ type Service struct {
 	UnikernelImages []string `json:"vm_images"`
 	Architectures   []string `json:"arch"`
 	Pid             int
+	OneShot			bool	 `json:"one_shot"`
```
melkodary, line 21: can we fix alignments? Is it tab indented or space or mixed?

melkodary (approved):
LGTM!

go_node_engine/model/Service.go
Malyuk-A: Nicely spotted.
Strange, on my end this looks like this:
image
I guess github did its thing

Is fixed

giobart:
It worked on my side as well at first

image

But after a while the state changed to failed

image

Then re-executed, swithced back to complete

image


I'm using the following deployment descriptor:
```
{
      "microservices" : [
        {
          "microserviceID": "",
          "microservice_name": "curl",
          "cmd": ["sh", "-c", "curl 10.30.55.55 ; sleep 5"],
          "memory": 100,
          "vcpus": 1,
          "vgpus": 0,
          "vtpus": 0,
          "bandwidth_in": 0,
          "bandwidth_out": 0,
          "storage": 0,
          "code": "docker.io/curlimages/curl:7.82.0",
          "state": "",
          "port": "",
          "added_files": [],
          "one_shot": true,
          "constraints":[]
        },
        {
          "microserviceID": "",
          "microservice_name": "nginx",
          "cmd": [],
          "memory": 100,
          "vcpus": 1,
          "vgpus": 0,
          "vtpus": 0,
          "bandwidth_in": 0,
          "bandwidth_out": 0,
          "storage": 0,
          "state": "",
          "port": "",
          "addresses": {
            "rr_ip": "10.30.55.55"
          },
          "added_files": [],
          "constraints": []
        }
      ]
}
```

Malyuk-A:
@giobart 
Thank you for spotting that completed jobs get turned into failed ones. I have not seen this before myself.
I was able to replicate your observations thanks to your provided SLA. 
I found the issue. The cluster-manager mongodb_client updates all jobs that are inactive for some time into failed jobs.
I have adjusted the code accordingly.
Now the issue is fixed.

giobart (approved):

giobart (approved):

PR #297, opened 2024-04-08
Updated startup scripts

giobart:
# What's new?

Added startup scripts for easy clusters and worker's installation

# Try it out: 

simply run: `curl -sfL oakestra.io/getstarted.sh | sh - `  to startup the latest stable release of 1-DOC
It will automatically be set up using local IPs and fetching the geo coordinates from your public address. 

More details and scripts under the `scripts/` folder

To test the locally always export the 
`export OAK_BRANCH=startup-scripts` env variable

# Tests:
tested in tum cluster with following setups:

- 1-doc
- separate 1 root and 1 cluster

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,54 +39,22 @@ In this get-started guide, we place everything on the same machine. More complex
 <a name="requirements"></a>
 
 - Linux machine with iptables
-- Docker + Docker Compose 
+- Docker + Docker Compose v2
 
 ### Your first cluster 🪵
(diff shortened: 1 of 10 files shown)


nitindermohan:
Tested on Mac OS 14.4.1 and Ubuntu 22.07. Works as expected. Approved!

-- Nitinder

nitindermohan (approved):

Issue #302, opened 2024-04-17
Update SLA documentation


giobart:
## Short
The doc file https://github.com/oakestra/oakestra/blob/develop/root_orchestrator/system-manager-python/sla/sla_documentation.md is outdated

## Proposal
Update this file with latest info such as
- Unikernel deployment 
- one-shot service
- cmd
- latest version
- env variables
- constraints type: `direct`is missing, while geo and latency must be marked as under development. 
- ips: `instances` and `closest` must be marked as under development
- connectivity: deprecated

Look the repo readme where these fields are currently documented. Move the description and examples from amin readme to this SLA doc document, add a link and shorten main README.md.

## Ratio
Outdated info

## Impact
Only docs

## Development time
9999 years

## Status
Looking for a volunteer

## Checklist
 
- [x] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

PR #309, opened 2024-04-17
307 allow larger service names

### LINKED ISSUE #307: Allow larger service names
## Short
Allow SLA containing larger service names

## Proposal 
refer to:  https://github.com/oakestra/oakestra-net/issues/168

## Solution 
Root orch validation and removal of max 8 char long service name limitation

## Status
development

## Checklist
- [x] Discussed
- [ ] Solved
- [ ] Tested

giobart:
fixes #307 

image

added regex-based name filter.

This allows alphanumeric application names up to 30 chars


TheDarkPyotr (approved):
LGTM 👍
Side note: to avoid losing track of the accepted format, we should update the documentation/format should be clearly detailed to the user during creation

giobart:
Good Point, I'll update the Readme as part of this PR, then we need to address the service format as part of this PR  https://github.com/oakestra/oakestra/issues/302
Additionally we can add info buttons in dashboard with service naming conventions and limitations, for this we need to open a new issue.

PR #313, opened 2024-04-24
Addons Engine + Marketplace

melkodary:
#315 

### PR:
Three major components created:
- **marketplace_manager**: is a server for hosting available addons.
- **addons_manager**: is a server, providing manipulation to addons database. User decide whether to run an addon or disable it by modifying the state of the addon in the database.
- **addons_monitor**: Polls the addons_manager and docker to achieve the target state of the addon. It is responsible for running, stopping and retrying addons. _Addons monitor utilizes docker outside docker: basically allows a container to manage containers of the host (outside) by mounting the socket volume. This approach is more safe/secure in_ comparison to dind.

#### Simple Run:
1. inside root_orchestrator: `docker compose up --build --force-recreate addons_monitor -d`
2. make a post request to `[POST] localhost:11102/api/v1/marketplace/addons`:

```
    {
      "name": "testing-addon",
      "services": [
        {
          "command": "sh -c 'while true; do echo \"Hello, World (testing!!) !\"; sleep 10; done'",
          "image": "alpine",
          "service_name": "my-alpine"
        }
   ]}
```

5. make another api request to `[POST] localhost:11101/api/v1/addons`:

```
    {
        "marketplace_id": (id from the response of the above request)
    }
```


addons_marketplace/marketplace-manager/api/v1/marketplace_blueprint.py
```diff
@@ -0,0 +1,34 @@
...
+    @marketplaceblp.arguments(MarketplaceFilterSchema, location="query")
+    @marketplaceblp.response(
+        200, MarketplaceAddonSchema(many=True), content_type="application/json"
+    )
+    def get(self, query={}):
+        return list(marketplace_db.find_approved_addons(query))
+
+    @marketplaceblp.arguments(MarketplaceAddonSchema, location="json")
+    @marketplaceblp.response(201, MarketplaceAddonSchema, content_type="application/json")
+    def post(self, data):
```
Malyuk-A, line 22: Is this method intended to add/post multiple addons at once or only single ones? If only single once then I think this method would be better placed at the singular "MarketPlaceAddonController" below not at this plural one.

addons_marketplace/marketplace-manager/api/v1/schema.py
```diff
@@ -0,0 +1,30 @@
...
+    ports = fields.Dict(keys=fields.String(), values=fields.String(), default={})
+    environment = fields.Dict(keys=fields.String(), values=fields.String(), default={})
+    labels = fields.Dict(keys=fields.String(), values=fields.String(), default={})
+
+
+class MarketplaceAddonSchema(Schema):
+    _id = fields.String()
+    name = fields.String(required=True)
+    description = fields.String()
+    services = fields.Nested(ServiceSchema, many=True, required=True)
```
Malyuk-A, line 25: What exactly are services in the context of marketplace addons?

addons_marketplace/marketplace-manager/services/marketplace_service.py
```diff
@@ -0,0 +1,38 @@
...
+import docker
+from db import marketplace_db
+
+
+def verify_addon(addon_id, addon):
+    logging.info(f"verifying addon-{addon_id}...")
+    client = docker.from_env()
+    for service in addon["services"]:
+        image_uri = service.get("image_uri")
+        try:
```
Malyuk-A, line 13: Do I understand this correctly?:
The verification is checking if the image can be pulled?

addons_marketplace/marketplace-manager/services/marketplace_service.py
```diff
@@ -0,0 +1,38 @@
...
+            logging.warning(f"Failed to pull image {image_uri}: {e}")
+            marketplace_db.update_addon(addon_id, {"status": "verification_failed"})
+            return
+
+    logging.info(f"Addon-{addon_id} verified")
+    marketplace_db.update_addon(addon_id, {"status": "approved"})
+
+
+def register_addon(addon):
+    addon = marketplace_db.create_addon({**addon, "status": "under_review"})
```
Malyuk-A, line 32: What reason is there to store addons in the DB that failed the verification? Is is due to the long asynch behavior and letting callers wait for the response?
(I.e. why not just store addons in the DB that passed verification?)

addons_marketplace/marketplace-manager/api/v1/marketplace_blueprint.py
Malyuk-A, line 1: Do you have some exemplary SLAs for this to try out locally? Maybe even add a simple example SLA as a README?

addons_marketplace/marketplace-manager/Dockerfile
```diff
@@ -0,0 +1,16 @@
...
+
+ADD . /
+
+ENV ADDON_MARKETPLACE_PORT=11102
+ENV ADDON_MARKETPLACE_MONGO_URL=localhost
+ENV ADDON_MARKETPLACE_MONGO_PORT=10007
+
+EXPOSE 11102
+
+CMD ["python", "addons_marketplace.py"]
```
Malyuk-A, line 16: image


```suggestion
CMD ["python", "marketplace_manager.py"]
```

addons_marketplace/marketplace-manager/marketplace_manager.py
```diff
@@ -0,0 +1,40 @@
...
+api.register_blueprint(marketplaceblp)
+
+
+@app.route("/", methods=["GET"])
+def health():
+    return "ok"
+
+
+if __name__ == "__main__":
+    app.run(host="0.0.0.0", port=ADDON_MARKETPLACE_PORT, debug=False)
```
Malyuk-A, line 40: Recommendation:

I am new to flask. During my own development I also stumbled upon this annoying warning:
image

I think to overcome this is pretty easy to do.

Have a look here how to use a "production ready WSGI server instead":
dependency needed: waitress = "3.0.0"

I use slightly different dependencies but the core thing can be resolved the same:

```python
from http import HTTPStatus

import flask_openapi3
import flops_manager.api.blueprints as blps
from flops_manager.api.utils.consts import FLOPS_MANAGER_PORT
from waitress import serve

_info = flask_openapi3.Info(title="FLOps Manager API", version="1.0.0")
app = flask_openapi3.OpenAPI(__name__, info=_info)


@app.route("/", methods=["GET"])
def health():
    return {"message": "ok"}, HTTPStatus.OK


def handle_api():
    for blp in blps.blueprints:
        app.register_api(blp)

    # Note (AFAIK): 'waitress' comes with its own logger.
    # We are already using the flops logger.
    # Thus we have duplicated logger outputs (not just for API things but generally).
    # To avoid this the waitress ones gets muted.
    # If something should be logged that concerns the API it will be logged via the flops logger.
    serve(app, host="0.0.0.0", port=FLOPS_MANAGER_PORT, _quiet=True)
```

addons_marketplace/marketplace-manager/db/mongodb_client.py
```diff
@@ -0,0 +1,22 @@
...
+mongo_marketplace = None
+app = None
+
+
+def mongo_init(flask_app):
+    global app, mongo_marketplace
+
+    app = flask_app
+
+    mongo_marketplace = PyMongo(app, uri=MONGO_ADDR_MARKETPLACE).db["marketplace"]
```
Malyuk-A, line 22: How exactly does this work?
If I am not mistaken there is no new MongoDB container running, will python spawn this on its own?

addons_marketplace/marketplace-manager/services/marketplace_service.py
Malyuk-A, line 1: Based on this diagram:

image

I expected to see a separate Addon Verifier component or even container. Is this your future vision of this and it is just for now a part of the marketplace-manager?

Malyuk-A (commented):
Ok this is my initial review round for the marketplace part.
I will continue with the addons engine now.

addons_engine/addons-manager/services/addons_runner.py
```diff
@@ -0,0 +1,291 @@
...
+        container.remove()
+
+    def _get_container_networks(self, container):
+        networks = container.attrs["NetworkSettings"]["Networks"]
+        return list(networks.keys())
+
+    def _get_container_ports(self, container):
+        return container.attrs["NetworkSettings"]["Ports"]
+
+    def maybe_create_networks(self, networks):
```
Malyuk-A, line 51: nitpick:

"maybe_create_networks" sounds strange to me. How about "fix_unavailable_networks" or "try_to_fix_unavailable_networks"

addons_engine/addons-manager/services/addons_runner.py
```diff
@@ -0,0 +1,291 @@
...
+DEFAULT_NETWORK_NAME = f"{DEFAULT_PROJECT_NAME}_default"
+
+ADDON_ID_LABEL = os.environ.get("ADDON_ID_LABEL") or "oak.addon.id"
+ADDON_MANAGER_LABEL = os.environ.get("ADDON_MANAGER_KEY") or "oak.plugin.manager.id"
+ADDON_SERVICE_NAME_LABEL = os.environ.get("ADDON_SERVICE_NAME_LABEL") or "oak.service.name"
+
+MARKETPLACE_API = f"{os.environ.get('MARKETPLACE_ADDR')}/api/v1/marketplace"
+
+
+class DockerAddonRunner:
```
Malyuk-A, line 21: Do you have some exemplary SLA/(test/dummy)addon I could try out to see if this works as intended?

addons_engine/addons-manager/tests/addons_service_test.py
```diff
@@ -0,0 +1,57 @@
...
+
+import docker
+import pytest
+import tests.container_utils as container_utils
+
+current_dir = Path(__file__).parent
+parent_dir = current_dir.parent
+sys.path.append(str(parent_dir))
+
+from services.addons_service import (  # noqa: E402
```
Malyuk-A, line 12: If I am not mistaken:
```suggestion
from services.addons_runner import (  # noqa: E402
```

addons_engine/addons-manager/tests/addons_service_test.py
Malyuk-A, line 1: image

Could it be that this const is not defined?

image

addons_engine/addons-monitor/addons_monitor.py
```diff
@@ -0,0 +1,40 @@
...
+        socketio.emit("get_manager_id")
+
+    socketio.on("connect", on_connect)
+    socketio.on("receive_manager_id", on_receive_manager_id)
+
+    # will keep trying to connect for 30 seconds
+    socketio.connect(ADDONS_MANAGER_ADDR, wait_timeout=30)
+
+    # wait until addon_manager is initialized
+    while not addon_manager:
```
Malyuk-A, line 32: Do you think it makes sense adding a sleep cmd here instead of just a pass? maybe this will be less "demanding" on the system?

Malyuk-A (commented):
Completed my first review round for the addons_engine 🙂

addons_engine/addons-manager/db/addons_db.py
```diff
@@ -0,0 +1,38 @@
...
+from bson.objectid import ObjectId
+
+
+def find_addons(filter={}):
+    return db.mongo_addons.find(filter)
+
+
+def find_addon_by_id(addon_id):
+    addon = list(find_addons({"_id": ObjectId(addon_id)}))
+    return addon[0] if addon else None
```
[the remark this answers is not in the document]
melkodary: `find_addons` is function that is used, so just to increase reusability I believe this way it's better.

addons_marketplace/marketplace-manager/api/v1/schema.py
melkodary: similar to docker-compose, the services keyword indicates the containers.

addons_marketplace/marketplace-manager/services/marketplace_service.py
melkodary: yes! inefficient but should be enough for now. I am working on a go implementation that does this more effieciently.

addons_marketplace/marketplace-manager/api/v1/marketplace_blueprint.py
melkodary: Sure!

addons_marketplace/marketplace-manager/api/v1/marketplace_blueprint.py
Malyuk-A: Ah yes I see. Ok makes sense

addons_marketplace/marketplace-manager/api/v1/schema.py
Malyuk-A: If you are fine with this that its ok. 
(For me who is not swimming in your topic I instantly associate this with oakestra services, so a simply change to addon_services would be cool, but up to you. I am fine with how it is now)

addons_marketplace/marketplace-manager/db/mongodb_client.py
melkodary: you'd need to spin up the mongo container yourself if you want to use it, hence:
```
MONGO_URL = os.environ.get("ADDON_MARKETPLACE_MONGO_URL")
MONGO_PORT = os.environ.get("ADDON_MARKETPLACE_MONGO_PORT")
```

However I could make it more clear that for local development you'd need to spin up your own mongo db container.

addons_marketplace/marketplace-manager/services/marketplace_service.py
melkodary: Thank you for pointing that out. . You are absolutely right and that's was my vision but for the time being I could indicate that it is just a process or something. I'll fix it.

addons_engine/start-up.sh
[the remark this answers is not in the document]
melkodary: addon-engine cannot be run as a container, otherwise installed containers are docker in a docker, defeating it's purpose of running containers in the control plane.

addons_marketplace/marketplace-manager/db/mongodb_client.py
Malyuk-A: This would be helpful, because all other oakestra things can be spun up somehow by compose files so this would reduce misconseptions/confusion

addons_engine/README.md
```diff
@@ -0,0 +1,14 @@
...
+
+## Ingoing Endpoints
+
+
+For more details checkout out the [openapi.json](./api/v1/openapi.json)
+
+
+## Start Service
+
+Run the service by running `./start-up.sh`
```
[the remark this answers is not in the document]
melkodary: Probably broke when I restructured some stuff. I'll fix it.

addons_engine/addons-manager/api/v1/addons_api.py
```diff
@@ -0,0 +1,83 @@
...
+@addonsblp.route("/")
+class AllAddonsController(MethodView):
+    @addonsblp.arguments(AddonFilterSchema, location="query")
+    @addonsblp.response(200, AddonSchema(many=True), content_type="application/json")
+    def get(self, query={}):
+        return list(addons_db.find_addons(query))
+
+    @addonsblp.arguments(AddonSchema, location="json")
+    @addonsblp.response(201, AddonSchema, content_type="application/json")
+    def post(self, addon_data):
```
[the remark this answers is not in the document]
melkodary: same answer.

addons_engine/start-up.sh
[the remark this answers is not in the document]
Malyuk-A: Isn't it possible to mount the docker.socket thus having access to the hosts docker engine and launch proper containers and not dind?

addons_engine/addons-manager/services/addons_runner.py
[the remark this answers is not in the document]
melkodary: I guess this could be cleaned up a little bit but 300 lines of code is a lot by which standard?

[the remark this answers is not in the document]
Malyuk-A: If you want to keep it as is ist is totally fine. I am simply used to python files that are < 120 LoCs 🙂

addons_engine/addons-manager/services/addons_runner.py
melkodary: Good point! I'll add some steps in the documentation for it.

addons_engine/addons-manager/tests/addons_service_test.py
melkodary: Oh yes! why does vscode not realize the refactor change 😠 Pycharm usually does.

addons_engine/addons-monitor/addons_monitor.py
melkodary: yeah I guess, it shouldn't loop forever I agree. I'll adjust.

addons_engine/addons-manager/tests/addons_service_test.py
Malyuk-A: Yeah, sometimes its a bit flaky 😅

addons_engine/start-up.sh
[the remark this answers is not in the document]
melkodary: I have not used this before, but it seems it's mostly configuration stuff so its something we could implement later. However if this is possible I'd be wary of the security concerns this would have, since containers should be totally isolated and don't have access to the host outside.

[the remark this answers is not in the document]
Malyuk-A: Fair points. Seems fine for now.

addons_marketplace/marketplace-manager/README.md
```diff
@@ -0,0 +1,48 @@
...
+For more details checkout out the [openapi.json](./api/v1/openapi.json)
+
+
+## Start Marketplace manager.
+
+Run the service by running 
+
+```
+$ docker build -t marketplace_manager .
+$ docker run marketplace_manager.py -p 11102:11102
```
Malyuk-A, line 16:
```suggestion
$ docker run -p 11102:11102 marketplace_manager
```

The original cmd does not work for me

addons_engine/README.md
Malyuk-A, line 12: I get the same error:
image

Even if I install vevns like this: sudo apt install python3-virtualenv

addons_engine/addons-manager/tests/__init__.py
Malyuk-A, line 1: The test works on my end now:
image

Malyuk-A (commented):
I had a second look

addons_engine/README.md
melkodary: Do any start-up.sh in the project work for you?

melkodary: I quickly searched the error you reported. Is it possible you don't have python3.8 installed in your machine? If so try modifiying the start-up.sh line to be:
`virtualenv --clear -p python3 .venv` (python3 instead of python3.8)

addons_marketplace/marketplace-manager/services/marketplace_service.py
melkodary: Currently I dont have a strong opinion about whether to keep it or remove failed ones. For the time being I dont see harm in keeping failed ones. Later if we find out that failed ones are no longer needed and they are taking a lot of space, we could setup a cron job the regularly cleans the db.

addons_engine/README.md
Malyuk-A: It works when I change pyhton3.8 to python3.
It would be cool if you could change that so that these scripts are not hardcoded to version 3.8 (I am using 3.10 and up)

addons_marketplace/marketplace-manager/services/marketplace_service.py
```diff
@@ -0,0 +1,46 @@
...
+def verify_addon(addon_id, addon):
+    logging.info(f"verifying addon-{addon_id}...")
+    client = docker.from_env()
+    for service in addon["services"]:
+        image = service.get("image")
+        image_id = None
+        try:
+            # TODO utilize opencontainers image spec to verify image instead of docker
+            logging.info(f"Pulling image: {image}")
+            pulled_image = client.images.pull(image)
```
giobart, line 17: If I'm not mistaken from the docker python SDK there is a command method `get_registry_data` that gives you the metadata from the image registry given an image name. No image pull requested. Maybe we can use that for this check?

giobart (changes requested):

addons_engine/addons-manager/addons_manager.py
```diff
@@ -0,0 +1,43 @@
...
+
+app.config["OPENAPI_VERSION"] = "3.0.2"
+app.config["API_TITLE"] = "Addon Manager Api"
+app.config["API_VERSION"] = "v1"
+app.config["OPENAPI_URL_PREFIX"] = "/docs"
+
+mongo_init(app)
+api = Api(app)
+
+# TODO remove this.
```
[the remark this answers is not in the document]
melkodary: I am modifying the logging configuration for testing purposes, making it printout everything even info.  So i just added a TODO for me to remove it before merging

addons_engine/addons-manager/services/addons_service.py
```diff
@@ -0,0 +1,57 @@
...
+MARKETPLACE_API = f"{MARKETPLACE_ADDR}/api/v1/marketplace/addons"
+
+addons_service = None
+
+
+# TODO: reuse this enum in addons_monitor
+class AddonStatusEnum(Enum):
+    # Allowed statuses for user to specify.
+    INSTALL = "install"
+    DISABLE = "disable"
```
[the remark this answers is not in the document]
melkodary: What I have in mind is that the user can issue either disable or install instructions, and when actually the addon is not running anymore because the `addons_monitor` stopped it, the status shown will be 'disabled'. Same case for the install. So when it's actually running it will be changed to active. 

I would love some feedback here because I have been having doubts about the different statuses and I believe it is still confusing somehow.

addons_engine/addons-monitor/utils/cleanup_handler.py
```diff
@@ -0,0 +1,11 @@
+from addons_runner import runner_types
+from services.monitor_service import addons_monitor
+
+
+def handle_shutdown():
```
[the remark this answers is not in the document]
melkodary: The addons_monitor.py when it is being shutdown. This is called when a SIGINT or SIGTERM is triggered.
```
def signal_handler(sig, frame):
    logging.info("Shutting down Addon Manager...")
    handle_shutdown()

    sys.exit(0)


if __name__ == "__main__":
    signal.signal(signal.SIGINT, signal_handler)
    signal.signal(signal.SIGTERM, signal_handler)
    try:
        addons_monitor.start_monitoring()  # This is a blocking call
    except Exception as e:
        logging.error("An error occurred while monitoring addons", exc_info=e)
```

addons_marketplace/marketplace-manager/services/marketplace_service.py
melkodary: Oh I see, that's even better. Thanks a lot :D

addons_engine/addons-manager/services/addons_service.py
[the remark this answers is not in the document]
melkodary: I could change it to, `installing` & `disabling`, indicating that it is not yet disabled or active.

addons_engine/addons-monitor/utils/cleanup_handler.py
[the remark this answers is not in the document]
melkodary: Just realized that this doesn't work when the process is running as a container.

giobart:
Can we migrate this implementation as compose override instead of directly in root compose? @melkodary

giobart (approved):

Issue #321, opened 2024-05-04
Custom Resources in root orchestrator


melkodary:
## Short
We want to be able to dynamically add new endpoints to new api objects and automatically handle their storage.

## Proposal
For further extending the system, this issue will handle dynamically creating new api objects, provide api endpoints for manipulating them and handle their storage.

Add the logic in resource abstractor to also benefit from the hooks.

**Flow:**
1. Adding a new resource should create a new endpoint at `/api/custom_resources/<new_resource>`. 
2. There will be a database called `custom_resources` that would hold the resources name along with their schema for validation - think of it as a collection of `metadata`.
3. Actual data of a resource should be in a collection of `custom_resources`.

## Ratio
This is useful for further extending the system.

## Impact
Resource abstractor

## Checklist
 
- [x] Discussed
- [ ] Documented
- [x] Implemented
- [x] Tested

oakestra-admin:
🥀 This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

oakestra-admin:
🥀 This issue is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

PR #322, opened 2024-05-04
Custom Resources in Root orchestrator

melkodary:
#321 

### How to use
1. create a custom resource which accepts an openapiv3 schema, e.g:
```json
{
    "resource_type": "employees",
    "schema":
    {
        "type": "object",
        "properties":
        {
            "age":
            {
                "type": "integer"
            },
            "address":
            {
                "type": "string"
            },
            "skills":
            {
                "type": "array",
                "items":
                {
                    "type": "string"
                }
            },
            "languages":
            {
                "type": "array",
                "items":
                {
                    "type": "string"
                }
            },
            "name":
            {
                "type": "string"
            },
            "email":
            {
                "type": "string"
            }
        }
    }
}
```
2. Now that a custom resource is created, you should notice that the under the custom_resources database there is new collection with the name `employees`. This also creates a new endpoint, `/api/v1/custom-resources/employees`. With the endpoints and storage ready, you can use RESTful api to manage the new entity.

### Future Work:
- Add new attribute to the schema called `plural_name` to enforce endpoints to always be in the form of `/api/v1/{plural_name}` thus following proper convention of naming uri.


giobart (changes requested):
Tested this branch and it looks like it introduces a regression in the front-end side, likely caused by slightly different objects returned by the API. 
It looks like service statuses and logs might have changed and the FE is not able to retrieve them. 

Screenshot 2024-07-29 at 13 40 21
Screenshot 2024-07-29 at 13 40 26

oakestra-admin:
🥀 This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

melkodary:
error is rooted in hooks branch, now fixed.

giobart (approved):
image


Tested and it works. I created an employee custom resoruce like the example you provided and it worked fine.

PR #334, opened 2024-05-23
instance number is an int

### LINKED ISSUE #331: Extremely Strange Deployment Behavior
## Short
This might be a combination of multiple things not just one.
When deploying services their deployment status seems to be wrong. 
Combined with multiple dashboard (GUI) errors, like missing logs, status, etc.

## Deeper description of the bug
Context
- This is not a flaky bug but it is persistent. (I.e. it shows up always for me).
- It shows up for me on my usual working machine, as well as on a secondary VM.
- These bugs seem to be independent of the Dashboard version. I have tried it out with latest as well as what I have been using for months (v0.4.205)
- These bugs only appeared after I merged the latest commits since the Accordion Release into my branch.
- I have observed these bugs on the unmodified last state of the develop branch
- I have rebooted my entire system several times, as well as completely pruned by docker images and volumes to make sure I use the latest components

I do not really understand why things are breaking so I will show you a comparison:

This is how it should look like: Notice the "RUNNING"
Description of Image

This is how it looks for me with the bugs:
Description of Image

---

This is how the service details should look like for the default nginx service:
Description of Image
Description of Image

This is how it looks for me with the bugs:
Description of Image
Description of Image

---

When inspecting the service details via the oak-cli or DB: this is how it looked like in the past:
Description of Image
Description of Image

This is how it looks like with the bugs:
Description of Image
Description of Image

Note: I think this was already a bit wrong in the past.
Because we have ~ these states:
Description of Image

So once a service is running (properly deployed) it should have the state "Running", but even in the past it had the state Node Scheduled ... (But in the Dashboard it properly showed the state Running ...)

---

Further interesting details:

The services are in fact still deployed. I can access them via ctr
Description of Image

Another extremely strange bug:
It seems to be the case that (at least in the GUI) all services are shared by all applications, i.e. the service list stays the same regardless of which app is selected. (Note: the "test1" app has just been added and is empty (i.e. it has no services). Even with non-empty apps their services get squashed together ...
Description of Image

## Solution 
I have no clue what the reason for this might be.

## Status
Currently, I cannot work on the lastest version of develop due to this bug, so I am reverting back to a stable state.
Looking for feedback: Has anyone else experienced this, or is my setup all of a sudden breaking?
Searching for a solution: Feel free to reach out to me to discuss this further. I think the root of this is due to the latest changes that took place in components I have not yet touched.

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

melkodary:
fix #331

### DIFF
--- a/root_orchestrator/resource-abstractor/api/v1/jobs_blueprint.py
+++ b/root_orchestrator/resource-abstractor/api/v1/jobs_blueprint.py
@@ -71,4 +71,6 @@ class JobInstanceController(MethodView):
         if ObjectId.is_valid(job_id) is False:
             raise exceptions.BadRequest()
 
-        return json.dumps(jobs_db.update_job_instance(job_id, instance_id, data), default=str)
+        result = jobs_db.update_job_instance(job_id, instance_id, data)
+
+        return json.dumps(result, default=str)
--- a/root_orchestrator/resource-abstractor/api/v1/jobs_blueprint.py
+++ b/root_orchestrator/resource-abstractor/api/v1/jobs_blueprint.py
@@ -72,5 +72,7 @@ class JobInstanceController(MethodView):
             raise exceptions.BadRequest()
 
         result = jobs_db.update_job_instance(job_id, instance_id, data)
+        if result is None:
+            raise exceptions.NotFound()
 
         return json.dumps(result, default=str)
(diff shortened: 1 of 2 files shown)


Malyuk-A:
Change seems simple enough. Right now one of our focuses is to simply fix the open issues to get the system working again.
So looks good to me.

giobart (approved):

PR #337, opened 2024-06-06
Statuses enhancement

Malyuk-A:
# Changes
- Adds a new library that can be imported anywhere.
- This library contains a "types" folder with a file for statuses.
- This file includes many important services, deployment, scheduling-related status
- It should be the single source of truth when it comes to these statuses - thus it is the place to add comments and explanations about when to use what status

# Affected Areas
- These changes only affect the Python code.
- I adjusted every occurrence of (non-HTTP) "status" I could find.

# Disclaimer
These changes are up for debate and are just the first step in bringing order into the chaos.
This MR should not become a huge unmergable beast, thus I want to merge it as soon as possible and use it as a ground for discussion for future improvements that build on top of it.

I have replaced all status strings with Enums however, to not break the running system (especially the HTTP responses and DB interactions, I often times use <status_enum>.value to get the original string value.
This means that there are several places left that should be refactored in follow-up MRs to ideally one day no longer depend on strings and fully support the enums. (E.g. I am using enums for my FLOps Plugin/Extension and because I use plain enums with pydantic I have no issues storing and loading them from/to the DB <- This would require a massive refactoring when applied to Oakestra).

I added many TODOs in the Code to be able to kick-start and follow up with discussions on how to improve things. There are simply too many open questions to resolve all of them at once.

This includes a color-coded Legend in the shared enum statuses file. We should Ideally make sure that we do not have any redundant status as well as have the same set of necessary statuses in Go and Python.

---

Here are a few figures to show what we currently are working with:

Original Mehdi Thesis |  Oakestra Paper 
:-------------------------:|:-------------------------:
  |  

Both are outdated

Go NodeEngine |  Latest 'COMPLETED' state addition
:-------------------------:|:-------------------------:
  |  

Both need polishing and unification.

# Sanity Check
I have always kept a close eye on the system and if it still works. I am able to create apps and services, deploy and delete them, i.e. the normal workflows seem to be working fine. I also keep an eye on the FE.


root_orchestrator/system-manager-python/requirements.txt
```diff
@@ -18,4 +18,7 @@ werkzeug==2.0.3
 MarkupSafe~=2.1.1
 flask_swagger_ui
 mongomock==4.1.2
-git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries/resource_abstractor_client
+#git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries/resource_abstractor_client
```
Malyuk-A, line 21: @Malyuk-A change the library dependencies to be pointing to the devel branch before merging

Malyuk-A:
I am looking for feedback on this, especially how I split up/grouped the statuses.

Feel free to suggest other splits and also to respond to open questions/TODOs I have added.
@giobart @TheDarkPyotr @melkodary

root_orchestrator/cloud-scheduler/requirements.txt
```diff
@@ -5,4 +5,7 @@ click~=8.1.7
 redis~=5.0.1
 requests~=2.27.1
 werkzeug==2.0.3
-git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries/resource_abstractor_client
+#git+https://github.com/oakestra/oakestra.git@develop#subdirectory=libraries/resource_abstractor_client
+git+https://github.com/oakestra/oakestra.git@service-status-enhancement#subdirectory=libraries/resource_abstractor_client
+git+https://github.com/oakestra/oakestra.git@service-status-enhancement#subdirectory=libraries/oakestra_utils_library
+icecream
```
melkodary, line 11: do we need icecream library?

melkodary (commented):
LGTM! Thank you for handling this status mess.

root_orchestrator/cloud-scheduler/requirements.txt
Malyuk-A: Good catch. I use it for debugging. So I will remove it before merging things. TODO @Malyuk-A

cluster_orchestrator/cluster-manager/mongodb_client.py
```diff
@@ -253,26 +268,30 @@ def mongo_update_jobs_status(TIME_INTERVAL):
...
+                )
+                if (
+                    job_is_inactive
+                    and job_status not in PositiveSchedulingStatus
+                    and job_status != DeploymentStatus.COMPLETE
+                ):
                     print("Job is inactive: " + str(job.get("job_name")))
-                    job["instance_list"][instance]["status"] = "FAILED"
-                    status = "FAILED"
+                    # TODO What is the difference between this status and the prev job_status ?
```
TheDarkPyotr, line 287: Just a minor clarification: what do you mean with _previous job_status_? I believe it's coherent if it update the job status from **Cluster Scheduled/Worker Scheduled** to **Failed** (_despite this should not happen considering the 'COMPLETED' state machine diagram_)

Malyuk-A: The previous job_status is simply the "other status" from a few lines above the "status" 
image

I had another look and managed to resolve the TODO. Now I simply create a new_job_status = FAILED and work with it.

TheDarkPyotr: You're right, more clear now

TheDarkPyotr:
Thanks for putting this together, looks great! 👍

libraries/oakestra_utils_library/oakestra_utils/types/statuses.py
```diff
@@ -0,0 +1,87 @@
...
+
+
+class DeploymentStatus(Status):
+    """Represents the state of a service during deployment,
+    after it got scheduled."""
+
+    CREATING = "CREATING"  # 🟪
+
+    RUNNING = "RUNNING"
+    # TODO: Figure out strong difference between ACTIVE and RUNNING states.
```
giobart, line 55: In principle:

NODE_SCHEDULED: Container not yet created but node is working on it
ACTIVE:  Container created but task not running yet. Therefore the image was downloaded, container metadata created, but task process inside the container is not running yet. E.g., because we're doing other pre-flight operations such as attaching the network. We can use a more conventional name such as "CREATED", I would personally prefer that. 
RUNNING: Container created and task running

libraries/oakestra_utils_library/oakestra_utils/types/statuses.py
```diff
@@ -0,0 +1,87 @@
...
+    # If there are no meaningful difference - merge both states into one.
+    # Else - adjust their names to emphasize the differences.
+    ACTIVE = "ACTIVE"  # 🟪
+
+    COMPLETED = "COMPLETED"  # 🟪
+
+    FAILED = "FAILED"  # 🟦 # TODO: Investigate the difference to DEAD
```
giobart, line 62: In theory, but as we know theory and practice don't always match:

FAILED: exit status != 0
DEAD: exist status == 0

root_orchestrator/system-manager-python/requirements.txt
giobart: If they manually point to develop, how do we make sure that stable branch points to main?

giobart (changes requested):

root_orchestrator/system-manager-python/requirements.txt
Malyuk-A: That is a great question.

This is currently already used by the Resource Abstractor Client Library. (@melkodary )
I currently see two solutions:
1) Having a dedicated MR for migrating things from devel to stable - in this MR the links need to be adjusted, this is not too pretty.
2) Publish these libraries as pip packages with proper versioning, same redundancy - "headaches" follow. 

I recommend discussing this in a maintainers meeting to come up with a nice convention.

libraries/oakestra_utils_library/oakestra_utils/types/statuses.py
Malyuk-A: Thank you for the clarification, I have adjusted the file accordingly.
I have renamed the ACTIVE status to CREATED.

libraries/oakestra_utils_library/oakestra_utils/types/statuses.py
Malyuk-A: Nice clarification. I added explanatory comments.

giobart:
Is it ready to be tested?

Malyuk-A:
> Is it ready to be tested?
@giobart 

I just tested it again locally and it works for me.
So feel free and test it.

Note. Before Merging I need to adjust the library imports to point to develop

giobart (approved):
image

LGTM!

root_orchestrator/system-manager-python/requirements.txt
melkodary: I believe the correct approach is the second solution proposed. However for the time being I think there is another solution that could be implemented - create the PR for library and merge it before merging the stuff of the main PR. This way no need to change the links everytime. In other words create two PRs: 
1. one for the new library or modification to the library (assuming changes don't break anything) and merge it.
2. Next create the main PR that uses this library.

PR #343, opened 2024-06-23
Add new clusters constraint

Malyuk-A:
Add a new constraint to allow specifying/restricting services to be deployed only on a specific set of clusters.

Why is this necessary:

This adds more flexibility and is a simple addition to the existing scheduling logic.

Concretely I need it to allow hierarchical FLOps, i.e. deploying learners for a specific cluster aggregator. I want to be sure that these learners are deployed on the same cluster as the cluster aggregator, plus I want to deploy one cluster aggregator per cluster. 

AFAIK we do not have any similar logic in place yet - the closest to it is to specify the node IP, which is too fine-grained for my use-case.

Here is a SLA you can use to test this out yourself:
```yaml
{
    "sla_version": "v2.0",
    "customerID": "Admin",
    "applications": [
        {
            "applicationID": "",
            "application_name": "clientsrvr",
            "application_namespace": "test",
            "application_desc": "Simple demo with curl client and Nginx server",
            "microservices": [
                {
                    "microserviceID": "",
                    "microservice_name": "curl2",
                    "cmd": [
                        "sh",
                        "-c",
                        "tail -f /dev/null"
                    ],
                    "memory": 100,
                    "vcpus": 1,
                    "vgpus": 0,
                    "vtpus": 0,
                    "bandwidth_in": 0,
                    "bandwidth_out": 0,
                    "storage": 0,
                    "code": "docker.io/curlimages/curl:7.82.0",
                    "state": "",
                    "port": "9080",
                    "added_files": [],
                    "constraints":[
                        {
                            "type":"clusters",
                            "allowed": [
                                "fictional_cluster"
                            ]
                        }
                    ]
                },
                {
                    "microserviceID": "",
                    "microservice_name": "nginx",
                    "cmd": [],
                    "memory": 100,
                    "vcpus": 1,
                    "vgpus": 0,
                    "vtpus": 0,
                    "bandwidth_in": 0,
                    "bandwidth_out": 0,
                    "storage": 0,
                    "state": "",
                    "port": "6080:80/tcp",
                    "addresses": {
                        "rr_ip": "10.30.30.30"
                    },
                    "added_files": [],
                    "constraints":[
                        {
                            "type":"clusters",
                            "allowed": [
                                "<YOUR CLUSTER NAME>"
                            ]
                        }
                    ]
                }
            ]
        }
    ]
}

```

### DIFF
--- a/cluster_orchestrator/cluster-scheduler/calculation.py
+++ b/cluster_orchestrator/cluster-scheduler/calculation.py
@@ -11,7 +11,12 @@ def calculate(app, job: dict) -> Union[dict, NegativeSchedulingStatus]:
 
     # check here if job has any user preferences, e.g. on a specific node, cpu architecture,
     constraints = job.get("constraints")
-    if constraints is not None and len(constraints) > 0:
+    if (
+        constraints is not None
+        and len(constraints) > 0
+        # The clusters constraint only plays a role in the cloud-scheduler.
+        and not (len(constraints) == 1 and constraints[0].get("type") == "clusters")
+    ):
         return constraint_based_scheduling(job, constraints)
     else:
         return greedy_load_balanced_algorithm(job=job)
--- a/cluster_orchestrator/cluster-scheduler/calculation.py
+++ b/cluster_orchestrator/cluster-scheduler/calculation.py
@@ -4,19 +4,18 @@ from typing import Union
 from mongodb_client import mongo_find_all_active_nodes
 from oakestra_utils.types.statuses import NegativeSchedulingStatus
 
+# TODO: Introduce proper constraint enums to oakestra-utils.
+SUPPORTED_CONSTRAINT_TYPES = ["latency", "geo", "addons"]
+
 
 def calculate(app, job: dict) -> Union[dict, NegativeSchedulingStatus]:
     print("calculating...")
     app.logger.info("calculating")
 
     # check here if job has any user preferences, e.g. on a specific node, cpu architecture,
-    constraints = job.get("constraints")
-    if (
-        constraints is not None
-        and len(constraints) > 0
-        # The clusters constraint only plays a role in the cloud-scheduler.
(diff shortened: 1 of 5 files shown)


cluster_orchestrator/cluster-scheduler/calculation.py
```diff
@@ -11,7 +11,12 @@ def calculate(app, job: dict) -> Union[dict, NegativeSchedulingStatus]:
 
     # check here if job has any user preferences, e.g. on a specific node, cpu architecture,
     constraints = job.get("constraints")
-    if constraints is not None and len(constraints) > 0:
+    if (
+        constraints is not None
+        and len(constraints) > 0
+        # The clusters constraint only plays a role in the cloud-scheduler.
+        and not (len(constraints) == 1 and constraints[0].get("type") == "clusters")
```
giobart, line 18: since constraints list is growing, can we find a slightly better way? E.g. supported_constraints constant list and then something like:

```
if set(constraints) & set(supported_constraints):
 constraint_based_scheduing(...)
```

giobart (changes requested):

cluster_orchestrator/cluster-scheduler/calculation.py
Malyuk-A: Good point. We need to extract and perform this check on the constraint types and not on the entire constraint dics, because one cannot convert them to a set easily AFAIK. 
image

Malyuk-A:
@giobart 
The scheduling bug is now fixed. 

I have successfully tested with the following permutations:
- no constraints
- just the cluster constraint (valid cluster name and non-existing cluster name)
- just the addon constraint (flops addon enabled & disabled on the worker)
- both - cluster & addon constraints (mix of the two mentioned above, one case where both are satisfied, only one is satisfied and none are satisfied)

giobart (approved):
Tested and it works

PR #363, opened 2024-09-24
Add JWT generator service and integrate it with the system manager

### LINKED ISSUE #218: Use separate asymmetric JWT token generator
## Short
A separate token generator generates all JWTs using asymmetric keys.

## Proposal
Currently the system manager uses a symmetric key, which is generated on each start of the system manager, to sign JWT access and refresh tokens via the JWTManager of the flask_jwt_extended library.
The proposed change would introduce a separate JWT token generator that generates these tokens using an asymmetric key pair which can be either provided or generated on start up and its public key can be retrieved from any service to validate issued tokens.
It would still keep the same flask_jwt_extended library but just forward the two token creation functions via http requests to the separate token generator service.

## Ratio
These are the key benefits of this change:
- Multiple services can issue tokens via the generator endpoints instead of only the system_manager (important for plugin components like multi-cloud broker).
- Multiple services can validate received tokens due to the availability of the public key
- Key retaining: Besides generating a key pair on start up a fixed key pair can also be provided as environment variables so that tokens can remain valid after restarts
- Separation of concerns: The signing key for JWTs is no longer owned by the system manager

## Impact
This affects only the root system manager key source for validating tokens and the token issuance functions.
The new generator service is added to docker compose along with the required env variables to the system manager.

## Development time
1 hour (as already implemented)

## Status
Discussed during multi-cloud meetings.
Generator implemented in separate repo by @michaelschlicker.
This repo to be forked, image pushed to own image registry and replaced in docker-compose before merging.

## Checklist
- [ X ] Discussed
- [ X ] Documented
- [ X ] Implemented
- [ ] Tested

giobart:
Fixes #218


giobart:
Few modifications required:
1: The comopse file images should point to the images we build 
2: extend this to 1-doc as well

giobart:
What's the outpcome when logging in using: <rootip>:10000/api/docs login procedure?

oakestra-admin:
🥀 This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

oakestra-admin:
🍂 This PR was closed because it has been stalled for 10 days with no activity.

giobart:
@melkodary is the login problem still there?

melkodary:
@giobart I will look into this next week. However if it's urgent, I would suggest that someone would takeover.

giobart:
Ok no problem keep me posted, if you're too busy just unassign yourself from the issue :)

melkodary:
@giobart, I Retested it, and it seems to work fine now. One thing, though, is I have never created a workflow for publishing the image before. Could you review what I did there?

oakestra-admin:
🥀 This PR is stale because it has been open 30 days with no activity. Remove stale label or comment or this will be closed in 10 days.

axiphi:
I would like to take a look at the PR

Malyuk-A:
@giobart, why did this issue remain open for so long  - AFAIK when it was written, there was already an implementation ready. Any idea?

image

image

https://github.com/michaelschlicker/simple-jwt-generator/blob/main/jwt_generator.py

melkodary:
@Malyuk-A it needs a review. Thats all.

PR #374, opened 2024-12-19
Fix install scripts

Malyuk-A:


go_node_engine/build/install.sh
```diff
@@ -61,4 +61,9 @@ sudo systemctl daemon-reload >/dev/null 2>&1
 sudo chmod 755 /bin/NodeEngine
 sudo chmod 755 /bin/nodeengined
 
-[ $? -eq 0 ] && echo "Done, installation successful" || echo "Installation failed, errors reported!"
```
Malyuk-A, line 64: Note that this line originally always returned exit code 0 - so the parent install script never reacted correctly even if this install failed

giobart (approved):

PR #381, opened 2025-01-25
Fix Typo breaking cluster images overrides

Malyuk-A:
### Context
I was trying out deploying the root and cluster separately with the modern/latest recommended curl scripts.
I noticed that the root worked fine but the cluster docker compose process failed due to a file it was not able to find.

image

After looking into this I figured out that the cluster-scheduler image was not properly replaced/pulled - but it should have been because we are using the image override yaml.

There is a very simple typo that breaks this.
As a result, the override file "exists" but is empty.

image


Malyuk-A:
@giobart I think we might need a new patch release

Malyuk-A:
I am very much looking forward to the automatic AWX testbed; it should really help us sanity check and detect such tiny mistakes that lead to breaking failures of fundamental/basic components/workflows.

Mjaethers (approved):
Oh dear, I believe this was me. Yeah the reason this slipped by was because I could test the script itself, but not the pulling of it was the image is only built post-merge. If the AWX Pipeline can check for this kind of thing, I'm also looking forward to it.
Thank you for looking into this!

giobart (approved):

PR #382, opened 2025-01-25
Add fixed default docker compose network name

Malyuk-A:
This MR adds a fixed docker (compose) network name to the 1-DOC and root-orchestrator compose files.

As a result, the docker network (of the control plane) will always have the same name, 'oakestra.'

image

Why do I welcome such a change?
The FLOps add-on communicates with the root manager. For simplicity's sake, it is hosted/deployed as a compose file currently attached to the same network as the already-running Oakestra control plane (1DOC or the root orchestrator compose).
While developing FLOps, I always used the multi-cluster approach, never the 1DOC; when I tried it with the 1DOC, FLOps failed because the network name was different.
By default, a docker-compose file's network name is based on the directory it resides in (e.g., cluster_orchestrator -> cluster_orchestrator_default).
AFAIK there are no native docker network aliasing, so getting FLOps to work for both is non-trivilal. 
This fix guarantees that the network name stays the same.

I successfully deployed the default application (curled from the curl service to the Nginx service) on the 1Doc and the multi-cluster one with these changes.


Malyuk-A:
@smnzlnsk, as our networKing, do you know of any better alternatives? Do these changes potentially interfere with or break certain Oakestra networking bits/features that I am unaware of?

smnzlnsk:
Hey Alex,
I am trying to understand why the docker network name is important to your application?

As this to my knowledge is a docker internal thing, that provides no external identification of network traffic. Maybe there is a way we can adjust your application instead, such that it becomes network-name agnostic?

Malyuk-A:
Great question @smnzlnsk,
From my understanding, each running Docker Compose suite creates its own isolated Docker network.
As far as I know, I cannot easily access services from Compose Suite A from Suite B; please correct me if I am wrong here.

So what I did is to tell the FLOps management compose suite to connect to the existing "external" network from the already running Oakesetra compose suite.

However, we have different names for those Oakestra networks.
When working with the install scripts - they create an "oakestra" folder in the user's home dir.
1) When running the 1-Doc install curl script - it will be launched from the ~/oakestra dir  -> thus the docker network name will be "oakestra-default"
2) When running the RootOrchestrator via the curl install script - it will be hosted in the ~/oakestra/root_orchestrator dir -> its network name will be root_orchestrator_default (vise versa for the cluster orchestrator - but I do not really care about that one because FLOps needs access to the Root Orchestrator)

During Local development when spinning up
3) 1-Doc from the repo .../run-a-cluster -> its network name will be "run-a-cluster_default"
4) The RO and CO will stay the same

I hope this makes it clearer.
I.e. I want to run the FLOps compose suite together with the Oakestra Control-Plane one.
An easy solution for this is to attach to the Oakestra docker network - but it could have different names - and there are no native docker network aliases (AFAIK)

smnzlnsk:
I see the problem now.
From my side there should be nothing speaking against this.

I guess having a static network name will only make it easier for future developers to extend the root orchestrator, since they will be able to hook up new components to the root network this way.

Approved!

smnzlnsk (approved):
Good one!

giobart (approved):

giobart (approved):

giobart (approved):

PR #389, opened 2025-02-17
added constraints file for ciffi version

### LINKED ISSUE #385: Build error: ERROR: Could not build wheels for cffi, which is required to install pyproject.toml-based projects
## Short
The GitHub action runner fails the build of the system manage and the jwt generator with the following error

## Proposal 
If we fix cffi version the problem should disappear. Apparently this was a known issue: https://github.com/python-cffi/cffi/issues/23#issuecomment-1845861410

## Status
Testing

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

giobart:
fixes #385

since cryptography is now forcing the update of ciffi to the buggy version 1.17.1 I added a constraints.txt file to force ciffi 1.15 as a workaround hoping it works

Locally it keeps building regardless, so I don't really know if this will work.


Malyuk-A (approved):
First time seeing constraints.txt - Let's try it out and hope for the best

PR #395, opened 2025-03-04
Fix dashboard container name

Mjaethers:
Changed the name of the container running the dashboard in the root orchestrator
oaekstra-frontend-container -> oakestra-frontend-container 

Just a small thing that was bugging me


giobart (approved):

giobart:
Thanks a lot for this PR, all checks went green! I took the liberty of including some override updates while in the process of updating the compose files. We can merge it :D

PR #398, opened 2025-04-09
Add Dockerfile to build a mongo container image that works on Raspberry Pi 4 Model B

### LINKED ISSUE #405: MongoDB container for root and cluster orchestrators is not compatible with Raspberry Pi 4 Model B
The MongoDB container version that Oakestra is using for its root- and cluster-orchestrators is not compatible with the Raspberry Pi 4 Model B (RPI-4b), as it requires a newer ARM processor variant. 
The standard MongoDB container for ARM requires an ARM version of ARMv8.2-A or higher, while the RPI-4b supports only ARMv8-A.

Since the Raspberry Pi is a popular SBC, it likely makes sense for Oakestra to support its second latest iteration (the only newer one is the Raspberry Pi 5).  

Oakestra could provide a compatible MongoDB container variant by building MongoDB with compilation flags that make it work on an RPI-4b.

axiphi:
## Overview
This PR adds a Dockerfile that builds a customized version of the standard mongo container image. This customized version uses compilation flags for mongo that allow it to run on a *Raspberry Pi 4 Model B*, which does not support the ARM version required by the standard mongo image. 

I put the new Dockerfile into a `hack` directory, which is a convention that stems from the Kubernetes project. This directory usually contains utility scripts for developers of a project, so it's suitable for the Dockerfile imo.

## Building
With `--build-arg MONGO_VERSION=8.0.4` and `--build-arg NINJA_JOBS=16`, building the image takes around 90 minutes on my machine.

## Testing
I confirmed that the custom built mongo binaries run on a *Raspberry Pi 400*, which has the same processor as the *Raspberry Pi 4 Model B*. I did not test the custom image in a complete Oakestra deployment, so it would be great if someone 

fixes #405


axiphi:
Closes #405.

giobart (changes requested):
Can you add a small readme in this hack folder to explain what this Dockerfile is for?

axiphi:
As I understand it, the lint failure is just a warning about the explicit `--platform` argument to the `FROM` statement of the Dockerfile. 
IMO it makes sense to keep it, as it prevents building invalid images when `--platform` is not set, as the binaries will only work on ARM so the platform has to be `linux/arm64/v8`.

giobart (approved):

PR #409, opened 2025-04-22
Use ruff for python linting & formatting

Malyuk-A:
resolves https://github.com/issues/assigned?issue=oakestra|oakestra|408


giobart (commented):
Why all these TODO(ME)?

Malyuk-A:
> Why all these TODO(ME)?

ruff uses a rule (that we can remove if we do not like it) that forces every TODO to be assigned a person/name.
i.e. TODO(ME) = TODO(Machmoud Elkordary) - I used the initials for faster typing and because we used this in HackMD all the time.

cluster_orchestrator/cluster-manager/mongodb_client.py
```diff
@@ -197,7 +197,7 @@ def mongo_aggregate_node_information(TIME_INTERVAL):
     mongo_update_jobs_status(TIME_INTERVAL)
     jobs = mongo_find_all_jobs()
 
-    # TODO: what is "more"?
+    # TODO(ME): what is "more"?
```
giobart:
```suggestion

```

giobart (changes requested):

Malyuk-A:
Once this is merged I will create the documentation update

giobart (approved):

PR #411, opened 2025-04-22
Fix cpu percentage

### LINKED ISSUE #410: Node total CPU usage not reported correctly
## Short
the gopsutils are reporting the load as avg number of processes running on cpu and not CPU%

## Proposal 
We can use cpu.Percentage module from gopsutils to calculate the CPU% variation across measurements.

## Solution 
We can use 
```
func getAvgCpuUsage() float64 {
	avg, err := cpu.Percent(0, false)
	if err != nil {
		return 100
	}
	if len(avg) == 0 {
		return 100
	}
	return avg[0]
}
```

instead of the current [getAvgCpuUsage](https://github.com/oakestra/oakestra/blob/a0d4e00a09065f0e8f332375b19d5204db93e1dc/go_node_engine/model/Node.go#L197) implementation. 

`cpu.Percent(0, false`) returns the CPU% aggregated for all CPUs calculated from cpu time variation of the last measurement when this method was called. This ensures non-blocking calls and a CPU% variation representative of the `MONITORING_CYCLE` defined in `nodeengined.go`

## Status
Implemented and tested

## Checklist
- [ ] Discussed
- [x] Solved
- [x] Tested

giobart:
fixes #410


Mjaethers (approved):
Looks fine to me!

PR #417, opened 2025-05-16
Fix containerd/api import which causes protobuf definition duplication.

axiphi:


axiphi:
The dependency versions used on develop already fixes this issue.

PR #425, opened 2025-06-05
Enable dynamic public ip updates

Mjaethers:
This will treat the Node IP Address as dynamic information which is regularly communicated to the cluster.
The user can choose whether the private or public IP should be transmitted with `sudo NodeEngine config network public` and `sudo NodeEngine config network private`, where private is the default.

If it's set to public, the NodeEngine will automatically find it's public address through an API call.

This is a pre-requisite for NAT Traversal.

### DIFF
--- a/cluster_orchestrator/cluster-manager/mongodb_client.py
+++ b/cluster_orchestrator/cluster-manager/mongodb_client.py
@@ -86,12 +86,15 @@ def mongo_find_node_by_id_and_update_cpu_mem(node_id, node_payload):
     # o = mongo.db.nodes.find_one({'_id': node_id})
     # print(o)
 
+    prev_ip = mongo_nodes.db.nodes.find_one({"_id": ObjectId(node_id)}).get("current_ip_address")
+
     time_now = datetime.now()
 
     mongo_nodes.db.nodes.find_one_and_update(
         {"_id": ObjectId(node_id)},
         {
             "$set": {
+                "current_ip_address": node_payload.get("ip", 0),
                 "current_cpu_percent": node_payload.get("cpu", 0),
                 "current_cpu_cores_free": node_payload.get("free_cores", 0),
                 "current_memory_percent": node_payload.get("memory", 0),
@@ -108,6 +111,10 @@ def mongo_find_node_by_id_and_update_cpu_mem(node_id, node_payload):
         },
         upsert=True,
     )
+
+    curr_ip = mongo_nodes.db.nodes.find_one({"_id": ObjectId(node_id)}).get("current_ip_address")
+    if prev_ip != curr_ip:
+        app.logger.info("IP_CHANGE - Node with id {0} changed its IP address".format(node_id))
 
     return 1
 
--- a/cluster_orchestrator/cluster-manager/mongodb_client.py
+++ b/cluster_orchestrator/cluster-manager/mongodb_client.py
@@ -85,12 +85,9 @@ def mongo_find_node_by_id_and_update_cpu_mem(node_id, node_payload):
     app.logger.info("MONGODB - update cpu and memory of worker node {0} ...".format(node_id))
     # o = mongo.db.nodes.find_one({'_id': node_id})
     # print(o)
-
-    prev_ip = mongo_nodes.db.nodes.find_one({"_id": ObjectId(node_id)}).get("current_ip_address")
-
     time_now = datetime.now()
 
-    mongo_nodes.db.nodes.find_one_and_update(
+    prev_document = mongo_nodes.db.nodes.find_one_and_update(
         {"_id": ObjectId(node_id)},
         {
             "$set": {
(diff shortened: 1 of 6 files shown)


Mjaethers:
@giobart I think perhaps this PR was forgotten, would you be able to take a look at it?

cluster_orchestrator/cluster-manager/mongodb_client.py
```diff
@@ -86,12 +86,15 @@ def mongo_find_node_by_id_and_update_cpu_mem(node_id, node_payload):
     # o = mongo.db.nodes.find_one({'_id': node_id})
     # print(o)
 
+    prev_ip = mongo_nodes.db.nodes.find_one({"_id": ObjectId(node_id)}).get("current_ip_address")
+
```
giobart: In order to avoid havin two mongo queries one next to the other can we use the return document option from the find_one_and_update below?
This is documented [here](https://pymongo.readthedocs.io/en/stable/api/pymongo/collection.html#pymongo.collection.Collection.find_one_and_update)

As far as I see, if from the update we use `return_document=ReturnDocument.BEFORE` parameter, the qiery will return the document before the update without your modifications, then you can compare the new ip with the old one. What do you think?

go_node_engine/cmd/conf.go
```diff
@@ -23,9 +23,13 @@ func init() {
 	setAuth.Flags().StringVarP(&certFile, "certFile", "c", "", "Path to certificate for TLS support")
 	setAuth.Flags().StringVarP(&keyFile, "keyFile", "k", "", "Path to key for TLS support")
 	setVirtualizationCmd.AddCommand(enableUnikernel)
+	setCni.AddCommand(explainNetManager)
 	setCni.AddCommand(enableNetwork)
 	setCni.AddCommand(disableNetwork)
 	setCni.AddCommand(enableManualNetwork)
+	setCni.AddCommand(explainIP)
+	setCni.AddCommand(publicIP)
+	setCni.AddCommand(privateIP)
```
giobart: Maybe I misunderstood, shouldn't this be part of explainIP command?
Like:
```suggestion
	explainIP.AddCommand(publicIP)
	explainIP.AddCommand(privateIP)
```

go_node_engine/cmd/conf.go
```diff
@@ -119,8 +123,12 @@ var (
 
 	// --- NETWORKING
 	setCni = &cobra.Command{
-		Use:   "network [auto/custom/off]",
-		Short: "Enable/Disable networking support",
+		Use:   "network [auto/custom/off] [public/private]",
+		Short: "Configure networking support",
+	}
+	explainNetManager = &cobra.Command{
```
giobart, line 129: What is the difference with the network command? Also, maybe the command name is missing from the Use section. 
E.g. for network `Use` is:
```
Use: "network [auto/custom/off] [public/private]",
```

go_node_engine/model/Node.go
```diff
@@ -17,16 +20,18 @@ import (
...
-	CONTAINER_RUNTIME RuntimeType = "docker"
-	UNIKERNEL_RUNTIME RuntimeType = "unikernel"
+// AddonType is the type of addon that the node supports
+var (
+	CONTAINER_RUNTIME config.RuntimeType = config.CONTAINER_RUNTIME
+	UNIKERNEL_RUNTIME config.RuntimeType = config.UNIKERNEL_RUNTIME
 )
 
-// AddonType is the type of addon that the node supports
+const SlowUpdateFactor = 60
```
giobart: Can you add a small comment to explain what this is?

giobart (changes requested):

cluster_orchestrator/cluster-manager/mongodb_client.py
Mjaethers: Yes, this makes sense. According to the documentation BEFORE is the default

go_node_engine/cmd/conf.go
Mjaethers: Originally I was trying to shoehorn public/private IP networking into the network command. But I think adding a new command called 'visibility' is cleaner. What do you think?

Mjaethers:
The config command now look like this:
```
Available Commands:
  addon          Enable/Disable addons
  applogs        Configure the log directory for the applications
  auth           Set Mqtt Authentication
  cluster        set the cluster address (and port)
  default        generates the default configuration file
  network        Configure networking support
  virtualization Enable/Disable a virtualization runtime support
  visibility     Use public or private IP
```
With visibility having the following format:
```
Usage:
  NodeEngine config visibility [command]

Available Commands:
  private     Disallow networking over the public IP address
  public      Allow networking over the public IP address

```

giobart:
What was the dependency loop that forced you to move the RUNTIME type to config?

go_node_engine/model/Node.go
```diff
@@ -145,6 +150,38 @@ func SetNodeId(id string) {
 }
 
 func getIp() string {
+	conf, err := config.GetConfFileManager().Get()
```
Mjaethers, line 153: This is why I refactored the Runtime Type into config. In order to check whether the IP should be public or private I need to read to configuration, so model has to import config, meaning that config can no longer import model so the RuntimeType has to be defined in config.

giobart (approved):

PR #429, opened 2025-06-25
Patch missing direct ocnstraint

giobart:
The direct constraint was missing from the cluster scheduler


Mjaethers (approved):
Looks good to me!

PR #431, opened 2025-07-02
Cleanup hanging containers

Mjaethers:
When the NodeEngine is still managing a service that no longer exists on the orchestration plane and the service was not correctly undeployed (e.g. if the root orchestrator was down) the NodeEngine will permanently report the error
```
ERROR-2025/07/02 23:01:32 ContainersManagement.go:392: Unable to fetch container task: no running task found: task Nginx.Test.curlv4.test.instance.1 not found: not found
ERROR-2025/07/02 23:01:32 ContainersManagement.go:392: Unable to fetch container task: no running task found: task Nginx.Test.nginx.test.instance.1 not found: not found
````
This also prevents a service of this name being deployed, resulting in phantom scheduling errors with the message
```
NoActiveClusterWithCapacity
```
To prevent this the NodeEngine should clean up any containers which it cannot find a task for


Mjaethers:
This branch was rebased onto the wrong one

PR #434, opened 2025-07-09
Fix Scheduler Constraints

Mjaethers:
The scheduler could not correctly deploy services when the SLA contained a constraint as there was a syntactic bug in the python code.
This fixes the issue and was tested on the edge testbed.

### DIFF
--- a/cluster_orchestrator/cluster-scheduler/calculation.py
+++ b/cluster_orchestrator/cluster-scheduler/calculation.py
@@ -83,7 +83,7 @@ def deploy_on_best_among_desired_nodes(job: dict, nodes) -> Union[dict, Negative
     else:
         desired_nodes_list = nodes.split(";")
         for node in active_nodes:
-            if node.get("node_info",{}).get("host") in desired_nodes_list:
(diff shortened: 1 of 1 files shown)


cluster_orchestrator/cluster-scheduler/calculation.py
```diff
@@ -83,7 +83,7 @@ def deploy_on_best_among_desired_nodes(job: dict, nodes) -> Union[dict, Negative
     else:
         desired_nodes_list = nodes.split(";")
         for node in active_nodes:
-            if node["node_info"]["host"] in desired_nodes_list:
+            if node.get("node_info").get("host") in desired_nodes_list:
```
giobart: if node_info is not available this still returns None right?

What about something like this?
```suggestion
            if node.get("node_info",{}).get("host") in desired_nodes_list:
```

giobart (changes requested):

cluster_orchestrator/cluster-scheduler/calculation.py
Mjaethers: Good catch! Thanks

giobart (approved):

PR #439, opened 2025-07-15
Refactor Resource Abstractor

### LINKED ISSUE #436: Unfiy Resource Names
## Short
Currently the ResourceAbstractor and the SLAs use different naming schemes for the same underlying resources (e.g "memory_in_mb" vs "memory"). As they refer to the same underlying object, we should be consistent in their naming.

## Proposal
Modify the names of the resources that the resource abstractor uses to be consistent with names used by the SLAs. This will allow existing SLAs to remain backwards compatible.

## Ratio
We should keep our naming consistent. This allows implementing a new scheduler with a single resourceList

## Impact
ResourceAbstractor

## Development time
1 Week

## Status
Discussed in the maintainers and with Giovanni

## Checklist
 
- [x] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

Mjaethers:
Fixes #436.
The names of resources within the resource abstractor now match the names of the constraints used by the SLAs.

This should not be merged until the scheduler has also been changed.


Mjaethers:
It makes more sense to close both Issue with one PR.
I have opened a new PR [here](https://github.com/oakestra/oakestra/pull/440)

PR #440, opened 2025-07-17
Abstract the Cloud Scheduler and Unify Resouce Names

### LINKED ISSUE #435: Abstract the Cloud Scheduler
## Short
Refactor the cloud scheduler so that it is structured to be more abstract in terms of underlying resource demands and scheduling algorithm.

## Proposal
It is possible to introduce a new scheduler to an Oakestra instance using the addons system, however this involves writing an entirely new scheduler as it is not easy to simply plug in a new algorithm.
The proposal is to abstract all scheduling operations in such a way to that is agnostic to the underlying scheduling algorithm and the resource constraints it considers.
To implement a new scheduler one would only have to define a new ResourceList; a collection of resource names (e.g. memory, vCPUs, ...) and their values, and a calculate function.

## Ratio
This will allow us to rapidly introduce new scheduling systems

## Impact
Cloud scheduler

## Development time
2 Weeks

## Status
Discussed in the context of my Bachelor Thesis

## Checklist
 
- [x] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

### LINKED ISSUE #436: Unfiy Resource Names

Mjaethers:
Closes #435.
Closes #436.

The Resource Abstractor refers to resources by the same naming scheme as the SLAs. The following names were changed:
- cpu_cores → vcpus
- memory_in_mb → memory
- gpu_cores → vgpus
- gpu_tot_mem → vram (v for video instead of virtual!)
- Remove “aggregated_” and “total_” prefix from resources where it is contextually obvious

The cloud scheduler has been abstracted. Most of the scheduling operations are now performed agnostically to the underlying scheduling algorithm, resource requirements and job (SLA) specifications.
To implement a new scheduler one has to implement the interfaces ResourceList and Algorithm (parametrised with ResourceList). The interfaces are defined in calculate/schedulers/interfaces.

Two schedulers were implemented:
1. rootBestCpuMemFit: Ranks clusters by available memory and cpu and then chooses the best one
2. rootRandom: Chooses a cluster at random provided it fits the constraints


root_orchestrator/scheduler/calculate/schedulers/interfaces/calculate.go
```diff
@@ -0,0 +1,51 @@
...
+	NoActiveClusterWithCapacity
+	NoWorkerCapacity
+	NoQualifiedWorkerFound
+	NoNodeFound
+)
+
+var NegativeSchedulingStatusName = map[NegativeSchedulingStatus]string{
+	TargetClusterNotFound:       "TargetClusterNotFound",
+	TargetClusterNotActive:      "TargetClusterNotActive",
+	TargetClusterNoCapacity:     "TargetClusterNoCapacity",
```
Mjaethers: Maybe this one can be deleted

Mjaethers:
The "total_" and "aggregated_" prefixes should be restored as they are important to unikernel scheduling

Mjaethers:
Add a Readme explaining the overall architecture and a readme in the calculate package expanding on implementing a new scheduler

giobart:
Would it make sense to move the scheduler as top level folder? This way this component can be built and imported by the cluster orch in the future without replication of code (once the resource abstraction will be implemented there as well ofc).

root_orchestrator/cloud-scheduler/Dockerfile
giobart, line 1: I guess the runner that builds the cloud-scheduler images inside the `.github` folder should be updated as well in this PR

root_orchestrator/scheduler/calculate/schedulers/rootBestCpuMemFit/calculate.go
```diff
@@ -0,0 +1,135 @@
...
+type BestCpuMemFit struct{}
+
+func (a BestCpuMemFit) ResourceList() []CpuMemResources {
+	var data []CpuMemResources
+	return data
+}
+
+func (a BestCpuMemFit) JobData() CpuMemResources {
+	var data CpuMemResources
+	return data
```
giobart: empty?

root_orchestrator/scheduler/calculate/schedulers/rootBestCpuMemFit/calculate.go
```diff
@@ -0,0 +1,135 @@
...
+
+	return nil
+}
+
+// BestCpuMemFit implements Algorithm
+type BestCpuMemFit struct{}
+
+func (a BestCpuMemFit) ResourceList() []CpuMemResources {
+	var data []CpuMemResources
+	return data
```
giobart: empty? What do you use ResourceList for?

root_orchestrator/scheduler/calculate/schedulers/rootBestCpuMemFit/calculate.go
giobart, line 1: can you write a test_calculate.go to test the Calculate function?

root_orchestrator/system-manager-python/blueprints/scheduling_blueprints.py
```diff
@@ -24,5 +26,9 @@ def post(self, *args, **kwargs):
         logging.log(logging.INFO, data)
         job_id = data.get("job_id")
         cluster_id = data.get("cluster_id")
+        if cluster_id is None:
+            # scheduling failed
+            status = data.get("status")
+            job_operations.update_job_status(job_id, convert_to_status(status))
```
giobart, line 32: shouldn't this be a failure status then?

giobart (changes requested):

root_orchestrator/scheduler/calculate/schedulers/rootBestCpuMemFit/calculate.go
Mjaethers: Yes, this is deliberate.
The scheduler can use this function to fetch an empty object of the appropriate ResourceList type. The scheduler can then unmarshal the job, which it receives as json, into the object, without prior knowledge on the specific ResourceList.

root_orchestrator/scheduler/calculate/schedulers/rootBestCpuMemFit/calculate.go
Mjaethers: Yes, this is deliberate.
The scheduler can use this function to fetch an empty object of the appropriate ResourceList type. The scheduler can then unmarshal the candidates, which it receives as json from the resource abstractor, into the object, without prior knowledge on the specific ResourceList.

root_orchestrator/system-manager-python/blueprints/scheduling_blueprints.py
Mjaethers: I'm not quite sure what you mean. If there is no cluster_id it means the scheduling failed and the status will be a NegativeSchedulingStatus.
Previously the scheduler directly called `job_operations.update_job_status`, but since this is a function from the python library it can no longer be called by the scheduler and has to be called by the system_manager.

Mjaethers:
> The "total_" and "aggregated_" prefixes should be restored as they are important to unikernel scheduling

Regarding this comment. The aggregation_by_architecture remains unchanged, so I don't think removing these prefixes makes a differences.
It definitely should be tested though. Does the oneshot pipeline deploy unikernels as well?

run-a-cluster/1-DOC.yaml
```diff
@@ -162,8 +162,8 @@ services:
   cloud_scheduler:
     image: ghcr.io/oakestra/oakestra/cloud-scheduler:latest
     pull_policy: always
-    container_name: cloud_scheduler
```
giobart, line 165: I guess the container name should remain cloud scheduler, otw how do we distinguish from the scheduler of the cluster orchestrator (if they are in the same machine?)

run-a-cluster/1-DOC.yaml
```diff
@@ -162,8 +162,8 @@ services:
   cloud_scheduler:
     image: ghcr.io/oakestra/oakestra/cloud-scheduler:latest
```
giobart: isn't the new image called `scheduler`according to `workflowsroot_scheduler_image.yml`?  For that workflow, would it make sense to keep the image name to cloud-scheduler?

scheduler/fig/scheduler-arch.drawio.svg
giobart, line 1: Can you open a PR in [alpha docs](https://github.com/oakestra/documentation/tree/main/content/version/alpha-v0.4.403) with updated scheduling algorithms updating the [original scheduling guide](https://www.oakestra.io/docs/concepts/scheduling/)?

giobart (changes requested):

run-a-cluster/1-DOC.yaml
Mjaethers: Yes, thank you very much

run-a-cluster/1-DOC.yaml
Mjaethers: Yes, good catch. I think this was a replace all error

scheduler/fig/scheduler-arch.drawio.svg
Mjaethers: Yes, it's here: https://github.com/oakestra/documentation/pull/74

giobart (approved):

PR #442, opened 2025-08-06
container deep clean

### LINKED ISSUE #430: Container deep cleaning
## Short
If containerd fails task cleanup, oakestra is not longer able to delete the dangling container. 

## Proposal 

Sometimes containerd can fail celaning up a container's task. With the current implementation, if that happens the removeContainer function of the NodeEngine fails to remove the container hosting the task. Additionally, we never remove the snapshots created by containerd snapshotters. 

We can update the removeContainer function from ContainerManagement.go as follows to obtain a deep cleanup. If a stage fails (e.g. task does not exists but the container does) then the cleanup just goes on undisturbed. 

```
func (r *ContainerRuntime) removeContainer(container containerd.Container) error {
	if container == nil {
		return nil
	}

	logger.InfoLogger().Printf("Clenaning up container: %s", container.ID())
	containerMetadata, err := container.Info(r.ctx)
	if err != nil {
		logger.ErrorLogger().Printf("Unable to fetch container metadata: %v", err)
	}

	//kill task
	task, err := container.Task(r.ctx, nil)
	if err == nil {
		err := killTask(r.ctx, task, container)
		if err != nil {
			logger.ErrorLogger().Printf("Unable to kill task: %v", err)
		}
	}

	//remove snapshotter
	if containerMetadata.Snapshotter != "" {
		currentsnapshotter := r.containerClient.SnapshotService(containerMetadata.Snapshotter)
		if currentsnapshotter != nil {
			err = currentsnapshotter.Remove(r.ctx, containerMetadata.SnapshotKey)
			if err != nil {
				logger.ErrorLogger().Printf("Unable to remove snapshotter %s: %v", containerMetadata.Snapshotter, err)
			}
			currentsnapshotter.Close()
		}
	}

	// remove container
	err = container.Delete(r.ctx)
	if err != nil {
		logger.ErrorLogger().Printf("Unable to delete container: %v", err)
	}

	return nil
}
```

## Status
looking for feedback

## Checklist
- [x] Discussed
- [ ] Solved
- [ ] Tested

giobart:
fixes #430


go_node_engine/virtualization/ContainersManagement.go
```diff
@@ -453,19 +453,43 @@ func (r *ContainerRuntime) forceContainerCleanup() {
 }
 
 func (r *ContainerRuntime) removeContainer(container containerd.Container) error {
+	if container == nil {
+		return nil
+	}
+
 	logger.InfoLogger().Printf("Clenaning up container: %s", container.ID())
```
Mjaethers:
```suggestion
	logger.InfoLogger().Printf("Cleaning up container: %s", container.ID())
```

Mjaethers (changes requested):

Mjaethers (approved):
Looks good to me!

Mjaethers:
Can this be merged? :)

PR #451, opened 2025-08-08
Fix resolv file and container monitoring

### LINKED ISSUE #449: Network resolution not working for networks with managed DNS
## Short
If you're in a network blocking 8.8.8.8 DNS the resolutions fails. 

## Proposal 
Managed networks sometims block requests to external DNS. 
We need to check first the managed systemd resolv file. If that's not available then we can default to the `8.8.8.8` DNS. 

## Solution 
Default to systemd managed resolv.conf file before user 8.8.8.8 

## Status
testing

## Checklist
- [ ] Discussed
- [x] Solved
- [ ] Tested

### LINKED ISSUE #450: When container is created the monitoring routine can kill it before task start
## Short
The monitoring routing might kill containers during startup

## Proposal 
In the Node Engine we have a monitoring routine killing containers without tasks. During container startup some time passes between container creation and task startup. If the monitoring routine falls between that interval it kills the container. 

## Solution 
Let the monitoring routine skip container deletion if the container startup just happened. 

## Status
testing

## Checklist
- [ ] Discussed
- [x] Solved
- [ ] Tested

giobart:
Fixes #449 
Fixes #450

### DIFF
--- a/go_node_engine/internal/daemon/nodeengined.go
+++ b/go_node_engine/internal/daemon/nodeengined.go
@@ -74,6 +74,10 @@ func main() {
 		}
 	}
 	if model.GetNodeInfo().Overlay {
+		logger.InfoLogger().Printf("Overlay network enabled 🟢")
+		// wait for systemctl to start the netmanager service
+		logger.InfoLogger().Printf("Waiting for NetManager to start...")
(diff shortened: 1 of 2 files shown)


Mjaethers (approved):
LGTM!

PR #455, opened 2025-10-01
Add crosvm runtime to NodeEngine

axiphi:
This PR adds a new runtime to NodeEngine that allows running instances as VMs with crosvm.

It changes how the different virtualization strategies (container, unikernel, etc.) are managed to not use global variables. 
This will affect runtimes still in development (WASM), which will need to be adjusted (though I think it should be pretty straightforward).

AI Summary:
This pull request introduces significant enhancements and refactoring to the Go Node Engine, focusing on improved runtime management, extensibility, and code clarity. The main changes include introducing a `RuntimeManager` abstraction for virtualization runtimes, updating MQTT handlers to utilize this manager, expanding model definitions, and updating dependencies. There are also several improvements to logging and documentation.

**Virtualization Runtime Management:**
- Introduced a `RuntimeManager` abstraction to manage different virtualization runtimes (e.g., Docker, Unikernel, CrosVM), replacing direct static calls with a more extensible and testable approach. All runtime-related code now uses this manager, including initialization in `main`, service monitoring, and runtime access in MQTT handlers. [[1]](diffhunk://#diff-bb70a5326ec3933d3d07ff1e14c0c5e045674ecfdd883062a7c3276564510a39L39-R55) [[2]](diffhunk://#diff-bb70a5326ec3933d3d07ff1e14c0c5e045674ecfdd883062a7c3276564510a39L88-R106) [[3]](diffhunk://#diff-760e2a64b2ba9bd14183beef606e032ee4bb87daf049a735fee360130b079876L10-R20) [[4]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L56-R63) [[5]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L69-R77) [[6]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L111-R127) [[7]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L122-R138) [[8]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L133-R150) [[9]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L142-R159)

**MQTT Handler Refactoring:**
- Refactored MQTT deploy and delete handlers to accept the `RuntimeManager` via a higher-order function, improving testability and reducing global state reliance. [[1]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L56-R63) [[2]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L69-R77) [[3]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L111-R127) [[4]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L122-R138) [[5]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L133-R150) [[6]](diffhunk://#diff-da95b39f5eae21d7c6ef7bae29b482142099eb49c32305c13b9f13ed28a63a60L142-R159)

**Model and Configuration Updates:**
- Added support for a new virtualization runtime, `crosvm`, across configuration and model layers. [[1]](diffhunk://#diff-fa51fc978f107eb31529ee991f368df3e082ba7614a9d670b46311c9936082cfR20) [[2]](diffhunk://#diff-58a094ea9d98c9e9934b2327cd2adf69129c185cefa51f32a2f0fe2d2cd94e6dL25-R34)
- Enhanced model structs with additional documentation and fields, including new fields for `storage` in `Service`, and expanded service status constants with comments for clarity. [[1]](diffhunk://#diff-180f38e14c937b0565aed45f7f21da202d388609450571aa9f0a9949c4e7d17aR19-R22) [[2]](diffhunk://#diff-180f38e14c937b0565aed45f7f21da202d388609450571aa9f0a9949c4e7d17aR44-R59) [[3]](diffhunk://#diff-58a094ea9d98c9e9934b2327cd2adf69129c185cefa51f32a2f0fe2d2cd94e6dR47)

**Dependency and Build System Updates:**
- Upgraded Go version to 1.24.0 and refreshed dependencies in `go.mod` for improved compatibility and security.
- Updated `.gitignore` to exclude built distribution files from third-party dependencies.

**Logging Improvements:**
- Refactored warning logger naming for consistency (`WarningLogger` → `WarnLogger`). [[1]](diffhunk://#diff-3c38b1e51b3b52760611a47ce6a78879a4240c4f0e9325b32d2a9276b95940caL11-R14) [[2]](diffhunk://#diff-3c38b1e51b3b52760611a47ce6a78879a4240c4f0e9325b32d2a9276b95940caR24-L38)

**Other Notable Changes:**
- Added `reexec.Init()` call at startup to support container-related tasks in sandboxed child processes, required by some OCI libraries.
- Minor improvements and clarifications in workflow YAML files for CI/CD. [[1]](diffhunk://#diff-4c3c4966694221feca4233e37a292d831cbc70017626d58adde74c139a8210f9L23-R24) [[2]](diffhunk://#diff-9dde0cfe11c07d54438f9b538c0283d3a84ae732d0c8b574d50d622b4f1bc11fL30-R30)

These changes collectively improve the maintainability, extensibility, and clarity of the codebase, especially in how different virtualization technologies are managed and integrated.


axiphi:
The new runtime is designed not to interfere with the existing NodeEngine functionality when the `crosvm` binary is not present on the worker-node system. Ideally, this behavior should be validated in a test cluster before this PR is merged.

go_node_engine/internal/daemon/nodeengined.go
```diff
@@ -16,13 +16,22 @@ import (
...
 const MONITORING_CYCLE = time.Second * 2
 
 var configs config.ConfFile
 
 func main() {
+	// The OCI libraries support performing certain container-related tasks in a sandboxed child process.
+	// For this to work, binaries using these libraries must use the reexec.Init() hook before doing anything else.
+	// This way, the library can take over execution in the mentioned child processes and will return true in that case.
+	if reexec.Init() {
+		return
```
giobart, line 32: is it intentional for it to just return without any log?

go_node_engine/util/iotools/remove.go
```diff
@@ -0,0 +1,27 @@
+package iotools
+
+import (
+	"go_node_engine/logger"
+
+	"github.com/spf13/afero"
+)
+
+func RemoveOrWarn(name string) {
```
giobart, line 9: not a bit fan of the names of the methods here. What if we just call them remove? the fact that it warns is fine anyway. Can you also give back the error if something fails?

go_node_engine/virtualization/internal/runtime/register.go
giobart, line 1: I like this but I'm not 100% convinced by the package name :D what about: rtregister or runtimeregister or something alond these lines?

go_node_engine/virtualization/internal/crosvm/internal/cloudinit/networkconfig.go
```diff
@@ -0,0 +1,33 @@
...
+	Version   int                              `json:"version"`
+	Ethernets map[string]NetworkConfigEthernet `json:"ethernets,omitempty"`
+}
+
+type NetworkConfigEthernet struct {
+	Match     *NetworkConfigMatch `json:"match,omitempty"`
+	Dhcp4     *bool               `json:"dhcp4,omitempty"`
+	Dhcp6     *bool               `json:"dhcp6,omitempty"`
+	Addresses []string            `json:"addresses,omitempty"`
+	// TODO(axiphi): gateway4 and gateway6 are deprecated, need to be exchanged
```
giobart, line 13: Don't forget to open an Issue about it

go_node_engine/virtualization/internal/crosvm/internal/image/containers.go
```diff
@@ -0,0 +1,260 @@
...
+		return err
+	}
+	defer iotools.CloseOrWarn(imageSource, "ImageSource "+id)
+
+	_, _, imageLayers, err := chooseImageParts(ctx, sys, imageSource)
+	if err != nil {
+		return err
+	}
+
+	// TODO(axiphi): check a special label in the manifest , so that only container images specifically built for this use-case work
```
giobart, line 74: What label? Is this something already implemented?

go_node_engine/virtualization/internal/crosvm/internal/image/containers.go
```diff
@@ -0,0 +1,260 @@
...
+
+	rootfsPath := path.Join(dstDirPath, internalRootfsFileName)
+	if err := fsimg.PackIntoSquashFsImg(rootfsDirPath, rootfsPath); err != nil {
+		return err
+	}
+
+	return nil
+}
+
+func chooseImageParts(
```
giobart, line 196: It's not clear to me what this function is for. Can you clarify?

go_node_engine/virtualization/internal/crosvm/internal/instance/network.go
```diff
@@ -0,0 +1,56 @@
...
+	AddressIpv4Cidr string
+	GatewayIpv4     string
+}
+
+func setupNetwork(service model.Service) (*networkConfig, error) {
+	if !model.GetNodeInfo().Overlay {
+		return nil, nil
+	}
+
+	if err := requests.CreateNetworkNamespaceForUnikernel(service.Sname, service.Instance, service.Ports); err != nil {
```
giobart, line 23: Do we need a PR in oakestra-net with unikernel network fixes for this to work or this works right away?

go_node_engine/virtualization/internal/crosvm/runtime.go
giobart, line 1: the crosvm runtime is inside the internal folder but the other runtimes are outside. Out of consistency, I guess this should we moved under /virtualization? Then, either we create folder for the each one of the other runtimes as well, or we make a CrossvmManagement.go inside /virtualization? I'm fine with both as long as all the runtimes look the same.

go_node_engine/virtualization/internal/networkutils/networkutils.go
giobart, line 1: maybe networkutils?

go_node_engine/virtualization/RuntimeDispatcher.go
giobart, line 1: I wonder, do we still need this RuntimeDispatcher? Can it be replaced with the allruntimes package? If that's the case, maybe the allruntime should be moved. Nothing that needs to be done in this PR but maybe can be discussed in an issue?

giobart (changes requested):

go_node_engine/internal/daemon/nodeengined.go
axiphi: Yes, the log wouldn't be visible anyway, as the reexec.Init() path is only hit when the `nodeengined` binary is specificially invoked with a special program name (in `os.Args[0]`) to run some child process logic instead of the normal `nodeengined` behavior. That is necessary because Go doesn't support forking.

go_node_engine/util/iotools/remove.go
axiphi: The functions I added wrap the standard `os.Remove` and `os.RemoveAll` for scenarios were there's no practical way to handle the errors that can occur from these operations. For example, when cleaning up temporary directories or while shutting down an Oakestra runtime, where aborting the process would make things even worse. 

The reason I named them differently is so that they are not confused with the standard `os.Remove` and `os.RemoveAll` functions and so that they are explicit about swallowing potential errors. Do you think that makes sense to you or do you prefer a different name anyway?

go_node_engine/virtualization/internal/runtime/register.go
axiphi: It actually doesn't even need to be a separate package, I think this was necessary at some point during development (to prevent circular imports), but isn't anymore.

I will move the register functionality directly into the `runtime` package.

go_node_engine/virtualization/internal/crosvm/internal/image/containers.go
axiphi: No, this is not yet implemented - I will add an issue about this.

The idea of the comment is to check the presence of a custom container image label which marks container images as compatible to be run as a VM. I will link the issue here for more details after I have created it.

go_node_engine/virtualization/internal/crosvm/internal/image/containers.go
axiphi: It selects the actual OCI image given a image reference, since those can point either to singular images or image indices - I will add a doc-comment.

go_node_engine/virtualization/internal/crosvm/internal/instance/network.go
axiphi: No, the current version in this PR is implemented to work with the current state of oakestra-net.

go_node_engine/virtualization/internal/crosvm/runtime.go
axiphi: I think the cleaner approach would be to move all runtime implementations into a sub-package of the `internal` folder. This way they are hidden from auto-complete in the rest of the codebase, which IMO makes sense, as most code should only use the generic `Runtime` interface.

I'm not sure if it makes sense to do this as part of this PR since it increases friction with existing code. If that's okay with you, I would create another issue about this. What do you think?

go_node_engine/virtualization/internal/networkutils/networkutils.go
axiphi: Fair point - will change it.

go_node_engine/virtualization/RuntimeDispatcher.go
axiphi: We do need to split the runtime functionality into two packages to prevent circular imports:

#### Current
```
RuntimeDispatcher/Manager -> All Runtime Impls -> Runtime Interface
         '-------------------------------------------^
```
#### Without Package Split (Circular)
```
 All Runtime Impls <-> Runtime Interface + RuntimeDispatcher/Manager
```

Since I moved the `allruntimes` into the `runtime` package though, I think this should already make the package structure less complex. What do you think?

go_node_engine/virtualization/internal/crosvm/internal/cloudinit/networkconfig.go
axiphi: Issue #467

go_node_engine/virtualization/internal/crosvm/internal/image/containers.go
axiphi: Issue #468

go_node_engine/virtualization/internal/crosvm/runtime.go
axiphi: Issue #469

giobart (approved):

giobart:
@axiphi except for a linting problem in the GO_MODULES we're good to go right?

axiphi:
@giobart I think so, yes. But do you see what the issue with GO_MODULES is? I can't really tell from the linter output

giobart:
I see this error: 

```
  level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"
  level=warning msg="[config_reader] The configuration option `output.uniq-by-line` is deprecated, please use `issues.uniq-by-line`"
  level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"
  level=warning msg="[config_reader] The configuration option `output.uniq-by-line` is deprecated, please use `issues.uniq-by-line`"
```

maybe an actual tidy will work?

giobart (approved):

giobart (approved):

.github/workflows/oneshot_testbed_approval.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         - name: Testbed Oneshot Execution - Waiting cycle
           uses: lewagon/wait-on-check-action@v1.3.4
           with:
-              ref: ${{ github.ref }}
+              ref: ${{ github.event.workflow_run.head_sha }}
```
giobart:
```suggestion
              ref: ${{ github.event.pull_request.head.sha }}
```

giobart (approved):

.github/workflows/oneshot_testbed_approval.yml
```diff
@@ -20,7 +20,7 @@ jobs:
         - name: Testbed Oneshot Execution - Waiting cycle
           uses: lewagon/wait-on-check-action@v1.3.4
           with:
-              ref: ${{ github.ref }}
+              ref: ${{ github.event.pull_request.head.sha }}
```
giobart:
```suggestion
              ref: ${{ github.sha }}
```

giobart (approved):

giobart (approved):

.github/workflows/oneshot_testbed_approval.yml
```diff
@@ -20,8 +20,8 @@ jobs:
         - name: Testbed Oneshot Execution - Waiting cycle
           uses: lewagon/wait-on-check-action@v1.3.4
           with:
-              ref: ${{ github.ref }}
-              check-name: 'Trigger Testbed Oneshot Execution'
+              ref: develop
+              check-name: Trigger Testbed Oneshot Execution ${{ github.event.pull_request.base.ref }}
```
giobart:
```suggestion
              check-name: Trigger Testbed Oneshot Execution ${{ github.event.pull_request.head.ref }}
```

giobart (approved):

PR #448, opened 2025-08-08
Refactor cluster orchestrator file structure

giobart:
Refactoring the structure of the cluster orchestrator folder

PR #456, opened 2025-10-07
Extend resource abstractor to cluster

### LINKED ISSUE #452: Extend the Resource Abstractor to the Cluster Level
## Short
The architecture of the scheduling system at the root level should be mirrored on the cluster level

## Proposal
At the root level the scheduling system has been modularised so that the available resources are abstracted by a component and another component can perform scheduling. This modularity also allow for [addons](https://www.oakestra.io/docs/manuals/extending-oakestra/setting-up/)
This system would also be extremely useful at the cluster level, especially considering the the extendability of the new [scheduler](https://github.com/oakestra/oakestra/pull/440)

## Ratio
This will allow addons and extendable scheduling at the cluster level

## Impact
Cluster Manager, Resource Abstractor

## Development time
2 Weeks

## Status
Discussed and under development

## Checklist
 
- [x] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

Mjaethers:
## Overview
Instead of managing and interfacing with mongodb directly the cluster now interacts with a resource abstractor instance.
Fixes #452 .

## Aggregation
The cluster performs job and worker aggregation as previously, but now adheres to the SLA naming conventions. The resources considered by the aggregator are now also divided into two categories:
1. Canonical Resources: These are resources the cluster always expects from the workers
2. Non-canonical Resources: These are custom resources provided by the node

### Aggregation Schemes
An aggregation scheme is a function, that accepts a worker, a default value and an accumulator. This function is applied to a worker and the accumulator (or default value) and returns an updated accumulator value. Once all the worker haves been processed, the accumulator value is propagated to the root.
The aggregation scheme is on a per-resource basis.

Canonical resources have their own hard coded aggregation scheme, for non-canonical resources the default aggregator is used which will process data types in the following way:
- numerical value -> sum 
- list -> flatten into single list
- other -> append to list

For *_percent values the aggregation scheme computes a running average.

## Rename cloud to root

Occurrences of the term cloud have been replaced with root for the sake of consistency.
This also affects oakestra-net components. The corresponding PR is here: https://github.com/oakestra/oakestra-net/pull/217

## Use same job_id everywhere

Previously jobs had an `_id` at the root level and a different `_id` at the cluster level. To track these jobs across both control planes, the jobs at the cluster level used a `system_job_id` corresponding to the `_id` at the root level. Since this introduced complexity without an obvious benefit the same `_id` is used for a job everywhere.
This also affects oakestra-net components. The corresponding PR is here: https://github.com/oakestra/oakestra-net/pull/217

This branch is rebased onto: https://github.com/oakestra/oakestra/pull/448


Mjaethers:
@giobart this is ready for review now :)

cluster_orchestrator/cluster-manager/blueprints/worker_blueprints.py
```diff
@@ -26,16 +26,20 @@ class ServiceController(MethodView):
         content_type="application/json",
     )
     def post(self):
-        logging.info("Incoming Request /api/node/register - to register node")
+        print("Incoming Request /api/node/register - to register node")
```
giobart: Why are we moving from the logging lib to print? I noticed this in all blueprints

cluster_orchestrator/cluster-manager/blueprints/service_blueprints.py
```diff
@@ -54,53 +56,60 @@ def post(self, system_job_id, instance_number):
...
+            job_id,
+            " instance ",
+            instance_number,
+            ". Result: ",
+            node_id,
+        )
+        if node_id is None:
+            # scheduling failed
+            status = data.get("status")
+            job_operations.update_job_status(job_id, convert_to_status(status))
```
giobart, line 100: We should send back a negative scheduling status right? Like NegativeSchedulingStatus.NO_WORKER_CAPACITY

cluster_orchestrator/cluster-manager/clients/resource_aggregation.py
giobart, line 1: The file is called worker_management but it seems like there is not much managmement but only aggregation operations. What about renaming it? What do you think?

cluster_orchestrator/cluster-manager/ext_requests/network_manager_requests.py
giobart, line 1: what is the difference between `net_plugin_requests.py` and `network_plugin.requests.py` ?

cluster_orchestrator/cluster-manager/ext_requests/scheduler_requests.py
```diff
@@ -0,0 +1,43 @@
...
+    request_addr = SCHEDULER_ADDR + "/api/calculate/deploy"
+    print(job)
+    try:
+        job["_id"] = job["_id"] + "/" + str(instance_number)
+        requests.post(request_addr, json=job)
+    except requests.exceptions.RequestException:
+        print("Calling scheduler", request_addr, "not successful.")
+
+
+def scheduler_request_replicate(job, replicas):
```
giobart: I believe rn the scheduler is only handling `/api/calculate/deploy` right? These replicate and status might come from old implementations. If that's the case, can we get rid of them?

cluster_orchestrator/cluster-manager/cluster_manager.py
```diff
@@ -85,15 +86,8 @@ def background_job_send_aggregated_information_to_sm():
             "time_interval": 2 * BACKGROUND_JOB_INTERVAL,
         },
     )
-    # job_dead_nodes
-    scheduler.add_job(
```
giobart, line 89: where is this handled now?

cluster_orchestrator/docker-compose.yml
```diff
@@ -82,7 +82,10 @@ services:
   # Cluster Manager
   cluster_manager:
     image: cluster_manager
-    build: cluster-manager/
+    build:
+      context: cluster-manager/
+      args:
+        - LIB_BRANCH=${LIB_BRANCH}
```
giobart: Does it mean that to run the compose file we alsways need to export LIB_BRANCH? What do you think about introducing a .env file with the default value to `develop` ?

root_orchestrator/system-manager-python/ext_requests/net_plugin_requests.py
```diff
@@ -20,9 +20,8 @@ def net_inform_service_deploy(job, job_id):
     request_addr = NET_PLUGIN_ADDR + "/api/net/service/deploy"
     logging.debug(request_addr)
 
-    job["_id"] = str(job_id)
-    job["system_job_id"] = str(job_id)
-    r = requests.post(request_addr, json={"deployment_descriptor": job, "system_job_id": job_id})
+    print("Inform service deploy with id ", job_id)
```
giobart: logging.info ?

root_orchestrator/system-manager-python/start-up.sh
```diff
@@ -9,11 +9,11 @@ source .venv/bin/activate
 export FLASK_ENV=development
 # export FLASK_DEBUG=False # TRUE for logging
 
-export CLOUD_MONGO_URL=3.120.37.66
-export CLOUD_MONGO_PORT=10007
+export ROOT_MONGO_URL=3.120.37.66
```
giobart: I know this file is quite old, but not that I noticed it, can we update the IP to localhost as default?

root_orchestrator/docker-compose.yml
```diff
@@ -10,7 +10,10 @@ services:
   # System Manager REST API of Root Orchestrator
   system_manager:
     image: system_manager
-    build: system-manager-python/
+    build:
+      context: system-manager-python/
+      args:
+        - LIB_BRANCH=${LIB_BRANCH}
```
giobart: same as the cluster's compose file, maybe a .env?

giobart (changes requested):

cluster_orchestrator/docker-compose.yml
giobart: During testing I had to fix this, I'm committing the solution right away

giobart:
For compatibility with existing `CLI` and `Dashboard` I added a cluster name mapping in the cluster blueprint

```
# Map candidate attributes to cluster attributes for compatibility with ext tools
# Deprecation note: this mapping should be removed when ext tools are updated to use
def map_cluster_attributes(x):
    x["cluster_name"] = x["candidate_name"]
    x["cluster_location"] = x["candidate_location"]
    return x

```

This maps the cluster_candidate and location to cluster_name and location. Current versions of CLI and dashbaord axpect these values. This can be deprecated later on.

cluster_orchestrator/cluster-manager/blueprints/worker_blueprints.py
Mjaethers: I couldn't get the python logging work together with the docker logging, meaning the logs were just lost. I can see if I can find a more elegant solution, potentially with a different python logging library

cluster_orchestrator/cluster-manager/blueprints/service_blueprints.py
Mjaethers: Yes, the scheduler has already encoded the appropriate scheduling error (since the cluster manager won't know *why* the scheduling failed) and that's why we simply pull the status out of the JSON field

cluster_orchestrator/cluster-manager/ext_requests/network_manager_requests.py
Mjaethers: Good catch, thanks. This must be a bug introduces when resolving merge errors

cluster_orchestrator/cluster-manager/ext_requests/scheduler_requests.py
Mjaethers: That's true, the scheduler doesn't handle this, but I wasn't sure if this might be re-implemented in the future so I wanted to leave this here. Maybe I can deprecate it with a comment?

I can also simply delete it if you think that's better

cluster_orchestrator/cluster-manager/cluster_manager.py
Mjaethers: I removed this, as it didn't appear to actually do anything.

This called:
```
def looking_for_dead_workers(interval):
    find_all_nodes()
```

Which in turn called
```
def find_all_nodes():
    global mongo_nodes
    return mongo_nodes.db.nodes.find()
```

Since the former doesn't propogate the retuned value I figured that this function didn't have a direct impact, aside from maybe some mongoDB caching, which we don't use any more.

cluster_orchestrator/docker-compose.yml
Mjaethers: Thanks!

root_orchestrator/system-manager-python/ext_requests/net_plugin_requests.py
Mjaethers: Same as above about the logging library

root_orchestrator/system-manager-python/start-up.sh
Mjaethers: That makes sense.

I'm just wondering if it's worth maintaining this file. The root orchestrator is normally deployed via docker (which our official documentation outlines), so perhaps we can just delete this and change the READMEs to only outline the docker deployment method?

Let me know what you think

Mjaethers:
I made sure to consistenly use the logging framework we prepare in `sm_logging.py` and `cm_logging.py` instead of the patchwork that didn't always cooperate with docker that we had previously

cluster_orchestrator/cluster-manager/ext_requests/scheduler_requests.py
giobart: Yes they are garbage, you can get rid of them! :D

cluster_orchestrator/cluster-manager/cluster_manager.py
giobart: Ah great! then if the functions `looking_for_dead_workers(interval)` is not used any longer as well we can delete it as well.

giobart: Maybe check when a node disconnects if the node is stil marked as INACTIVE

root_orchestrator/system-manager-python/start-up.sh
giobart: RN let's jsut replace 3.120.37.66 with 172.17.0.0

giobart:
I see this issue:

```
  File "/home/cm/Documents/oak-developement/oakestra-452-extend-resource-abstractor-to-cluster/cluster_orchestrator/cluster-manager/clients/resource_aggregation.py", line 83, in aggregate_workers
    for w in workers:
TypeError: 'NoneType' object is not iterable
```

giobart:
the .vscode Debug Cluster Scheduler profile should be updated

Mjaethers:
So I checked if this was decrementing "active nodes" when a worker disconnects and it wasn't, so thanks for spotting it. I went back to develop to see how it worked there, couldn't figure it out, tested it; it turns out the develop branch was also failing to mark disconnected nodes as inactive.

Long story short: This now works. In fixing this I noticed a few other bugs to do with the aggregation and patched them, so this is ready for review again :)

Mjaethers:
@giobart any news on this PR? :)

giobart (approved):

PR #457, opened 2025-10-15
Disable CGO

giobart:
We should disable CGO to increase the compatibility and portability


giobart:
The build passes because we don't have dependencies requiring CGO btw

Mjaethers (approved):
Yep, makes sense. This looks good to me

PR #465, opened 2026-01-21
Addons Marketplace Dashbaord

### LINKED ISSUE #464: Addons Marketplace Dashbaord
## Short
It would be nice to have a small dashbaord to handle custom resources, addons and the marketplace in your oakestra cluster. 

## Proposal
- The dashboard is meant to be exposed only inside Oakestra, not as public dashboard.  
- It must contain sections for Marketplace, Installed Addons, Hooks management, Custom Resources management
  - Marketplace:
    -  List of addons and addons creation menu
  - Installed Addons: list of installed addons, with possibility to disable and uninstall them
  - Hooks management: List of available hooks and an option to create one for each event type.
  - Custom Resources: 
    - List of custom reources definition and instances
    - Add custom resource definition
    - Add custom resource instance
- The dashbaord can connect to a local marketplace but a public marketplace should be configurable as well
- The dashboard must prioritize ease of use and lightweight operations over design as this is for infrastructure managers only. 

## Ratio
Easy navigation and management of the Custom resources, hooks and Addons 

## Impact
Root orchestrator, resource abstractor, addons manager

## Development time
2 Weeks

## Status
development

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

giobart:
Fixes #464 

image


addons_engine/addons-dashboard/README.md
```diff
@@ -0,0 +1,337 @@
...
+│   │   └── app.routes.ts        # Routing (if needed)
+│   ├── index.html               # HTML entry point
+│   ├── main.ts                  # Angular bootstrap
+│   └── styles.css               # Global styles
+├── angular.json                 # Angular configuration
+├── tsconfig.json                # TypeScript configuration
+├── package.json                 # Dependencies
+├── Dockerfile                   # Docker build
+└── nginx.conf                   # nginx configuration
+
```
HMF2475, line 315: Here should be a ``` that's missing

addons_engine/addons-dashboard/src/app/components/marketplace/marketplace.component.ts
HMF2475, line 1: I really like the addition of the Manual Mode! However, I feel the current implementation isn't quite abstract enough to distinguish it from the JSON mode yet. Since users still have to write the services array in raw JSON.

I think it would be a good UX improvement to convert the services section into a UI-driven list. It would be a lot easier pressing a button "Add Service" and fill in standard inputs (like Name, Image, etc.) without worrying about JSON syntax at all.

Additionally, I noticed that Manual Mode is currently missing the configuration for volumes and networks. In JSON mode, users can define drivers and options for these resources, but it isn't possible in Manual Mode .

addons_engine/addons-dashboard/src/app/components/custom-resources/custom-resources.component.ts
```diff
@@ -0,0 +1,299 @@
...
+    this.newDefinition = { resource_type: '', schema: {} };
+    this.schemaJson = '';
+  }
+
+  submitDefinition(): void {
+    try {
+      this.newDefinition.schema = JSON.parse(this.schemaJson);
+      this.resourceAbstractorService.createCustomResource(this.newDefinition).subscribe({
+        next: () => {
+          alert('✅ Resource definition created successfully!');
```
HMF2475: I noticed we are using native JavaScript alert() calls for success and error messages.

It isn't neccesarily a bad thing but maybe to maintain consistency and provide a non-blocking UX, would be better use a Notification Service like the one already [implemented for the dashboard](https://github.com/oakestra/dashboard/blob/develop/src/app/shared/modules/notification/notification.service.ts)

addons_engine/addons-dashboard/src/app/services/config.service.ts
```diff
@@ -0,0 +1,55 @@
...
+
+    this.initialized = true;
+  }
+
+  saveConfig(config: AppConfig): void {
+    this.config = config;
+    localStorage.setItem(this.CONFIG_KEY, JSON.stringify(config));
+  }
+
+  resetToDefaults(): void {
```
HMF2475: This method should be async and await this.loadConfig(). Without this, the method resolves before the configuration is fully reloaded and could lead to race conditions

addons_engine/addons-dashboard/src/app/services/resource-abstractor.service.ts
```diff
@@ -0,0 +1,90 @@
...
+  deleteCustomResource(resourceType: string): Observable<void> {
+    return this.http.delete<void>(`${this.baseUrl}/custom-resources/${resourceType}`);
+  }
+
+  getResourcesByType(resourceType: string, filters?: { [key: string]: any }): Observable<any[]> {
+    let url = `${this.baseUrl}/custom-resources/${resourceType}`;
+    
+    // Add query parameters if filters are provided
+    if (filters && Object.keys(filters).length > 0) {
+      const params = new URLSearchParams();
```
HMF2475: For consistency with the other services (like marketplace.service.ts) and to follow Angular best practices, we should use HttpParams here as well. 
It should look something like this:
```TypeScript
getResourcesByType(resourceType: string, filters?: { [key: string]: any }): Observable<any[]> {
  const url = `${this.baseUrl}/custom-resources/${resourceType}`;
  
  let params = new HttpParams();
  
  if (filters) {
    Object.keys(filters).forEach(key => {
      if (filters[key] !== null && filters[key] !== undefined && filters[key] !== '') {
        params = params.set(key, String(filters[key]));
      }
    });
  }
  
  return this.http.get<any[]>(url, { params });
}
```

addons_engine/addons-dashboard/src/app/components/installed-addons/installed-addons.component.html
```diff
@@ -0,0 +1,79 @@
...
+    <div class="meta">
+      <p><strong>ID:</strong> <code>{{ addon._id }}</code></p>
+    </div>
+    <div class="actions">
+      <button class="btn btn-primary btn-small" (click)="viewDetails(addon)">View Details</button>
+
+      @if (addon.status?.toString()?.toLowerCase()?.trim() === 'disabled') {
+        <button class="btn btn-danger btn-small" (click)="uninstallAddon(addon._id)">  Uninstall </button>
+      } @else {
+        <button class="btn btn-danger btn-small" (click)="uninstallAddon(addon._id)">  Disable </button>
```
HMF2475: I saw that that clicking "Disable" triggers the "Uninstall" confirmation alert, which is confusing.

Even though the backend correctly handles the first DELETE request as a "disable" action, the UI tells that you're going to uninstall it. 

It would be better to show a distinct confirmation message for disabling (e.g., 'Are you sure you want to disable?') to avoid misunderstandings.

HMF2475 (commented):
I really liked this Addons Marketplace implementation! It seems like a significant quality-of-life improvement and the architecture is generally solid.
I did find a few edge cases and UI issues that we may address before merging. I've left some notes in the files.

Mjaethers (commented):
I really like this!

My only gripe is that many statuses don't refresh themselves, i.e. I have to cycle tabs in order to see when something is finished installing, or do uninstall an addon after deleting it.

giobart:
Good point @Mjaethers, what about adding a refresh button? Would that work?

giobart:
@Mjaethers @HMF2475 In the last couple of commit I did the followinf:
- fixed a bug from #456 that prevented scaling of instances at cluster level, it was caused by a patch method in the resource abstractor instead of a put for new job instances 
- Added the marketplace and addons as default in Oakestra. They can be turned off with a new override `override-no-addons.yml`

If you agree with this last change, I guess we need to open two issues

(1) oakestra/documentation: document the new override and change docs related to marketplace (add dashbaord and setup)
(2) oakestra/dashboard: add a menu option, only for system Admin and if ROOT_URL:11103 is reachable from current network to open in a new tab the Marketplace UI. 

What do you think?

[Edit]
Based on the discussion in the last Maintainer's meeting:
- we keep the addons system active by default and merge this PR
- we add the link option in the dashboard as described by (2)
- we create an issue to fully embed this marketplace-dashboard in the official oakestra dashbaord (they are both in Angular)
- we integrate hooks and addons management in oak_cli
- we authenticate the addons API like root APIs via token

PR #466, opened 2026-02-03
Enable addons at cluster

### LINKED ISSUE #462: Enable Addons System at Cluster
## Short
The addons system that is currently functional for the root orchestrator should be extended to the cluster orchestrator.

## Proposal
Extending the resource abstractor to the cluster (#456) paves the way for extending the addons system to the cluster. Ideally a single marketplace manager at the root should suffice, while a addons manager and monitor can be deployed with a cluster orchestrator.

## Ratio
The addons system is a powerful paradigm of the Oakestra framework and there are many uses cases for it at the cluster.

## Impact
Cluster Orchestrator, Resource Abstractor, Addons Manager/Monitor

## Development time
1 week

## Status
Looking for feedback

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

Mjaethers:
Closes #462.

This PR leverages PR #456 to extend the addons system to the cluster.

This PR is rebased onto #465


Mjaethers:
I had to implement a naming scheme for the containers (i.e. root-<addon-name> or cluster-<addon-name>) so that the same addon could be deployed at the root and cluster if both are running on the same machine.

I also fixed a small bug with the logs not being preserved after a container fails

addons_engine/addons-manager/addons_manager.py
```diff
@@ -55,4 +55,4 @@ def health():
 
 
 if __name__ == "__main__":
-    app.run(host="::", port=ADDONS_MANAGER_PORT, debug=False)
+    app.run(host="0.0.0.0", port=ADDONS_MANAGER_PORT, debug=False)
```
giobart: 0.0.0.0 will remove ipv6 compatibility

addons_engine/addons-monitor/services/monitor_service.py
```diff
@@ -226,29 +229,44 @@ def update_addon(self, addon_id, data):
...
         elif not self._failed_containers.get(addon_id, None):
+            new_count = curr_retries + 1
             logging.info(
-                f"Addon-{addon_id}: container '{container.name}' exited with code {exit_code}"
+                f"Addon-{addon_id}: container '{container.name}' exited with code {exit_code} "
+                f"(retry {new_count}/{MAX_CONTAINER_RETRIES})"
             )
-            self._retry_containers[addon_id][container.id] = curr_retries + 1
+            self._retry_counts[addon_id][container.id] = new_count
+            self._pending_restart[addon_id].add(container.id)
```
giobart, line 273: Why do we need both `_prending_restart` and `_retry_counts` ?

cluster_orchestrator/override-no-addons.yml
giobart, line 1: I think we also need to update the following overrides as well:
- https://github.com/oakestra/oakestra/blob/develop/cluster_orchestrator/override-ipv6-enabled.yml
- https://github.com/oakestra/oakestra/blob/develop/cluster_orchestrator/override-images-only.yml
- https://github.com/oakestra/oakestra/blob/develop/cluster_orchestrator/override-network-host.yml

Maybe we don't need 1DOC updates as that's a 1 cluster config anyway so we don't need root and clsuter plugin manager.

resource-abstractor/db/custom_resources_db.py
```diff
@@ -59,6 +59,9 @@ def update_resource(resource_type, id, data):
     # Remove _id from data to avoid MongoDB immutable field error
     update_data = {k: v for k, v in data.items() if k != "_id"}
 
+    # Remove _id from data to avoid MongoDB immutable field error
+    update_data = {k: v for k, v in data.items() if k != "_id"}
```
giobart: Is this a duplicate line with the line above?


```suggestion
```

root_orchestrator/override-no-addons.yml
giobart, line 1: Probably you need to update this override as well: 
- https://github.com/oakestra/oakestra/blob/develop/root_orchestrator/override-network-host.yml

giobart (changes requested):

addons_engine/addons-monitor/services/monitor_service.py
Mjaethers: Yes, I think so.
`_pending_restart` is for tracking containers that we are going to try deploying again. `_retry_counts` is for tracking that we haven't exceeded `MAX_CONTAINER_RETRIES`, if we have we add the container to `_failed_containers`.

This fixes a bug whereby we wouldn't remove a container from `_retry_containers`, meaning we would keep restarting it, even if it now would have recovered (until we hit max_retries). Now we wait until it's restarted before checking again.

resource-abstractor/db/custom_resources_db.py
Mjaethers: Yeah, this must be a marge error

addons_engine/addons-manager/addons_manager.py
Mjaethers: Yeah, I added this because I mis-diagnosed a bug and forgot to remove it. Thanks!

Mjaethers:
@giobart I fixed the small bugs and adapted the override compose files

resource-abstractor/db/custom_resources_db.py
```diff
@@ -59,6 +59,8 @@ def update_resource(resource_type, id, data):
     # Remove _id from data to avoid MongoDB immutable field error
     update_data = {k: v for k, v in data.items() if k != "_id"}
 
+    # Remove _id from data to avoid MongoDB immutable field error
+
```
giobart:
```suggestion
```

giobart (changes requested):

giobart (approved):

PR #478, opened 2026-02-22
Fix auto re-deployment routine

### LINKED ISSUE #474: Auto re-deployment routine not working
## Short
The auto-redeployment of failed jobs in a cluster is not working any longer in `develop`

## Proposal 
Example: 

Image

The DEAD job should be redeployed by the cluster but in my case it remained like this for a while

## Solution 
Check cluster jobs for monitoring and re-deployment

## Status
looking for feedback, searching for a solution

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

Mjaethers:
Fixes #474 

The bug consists of two components:
1. Dead service instances were not being re-deployed
2. The status was not being set correctly

The fixes:
1. Simple one liner that prevented the cluster manager from deleting a job when it's only service was being re-deplyoed
2. The statuses are now being updated more thoroughly so when observing the job statuses directly in the mongo db they are being set correctly, however since the service is being stopped and started quite rapidly, the update window to the root is too small. The result is that at the root (i.e. via the CLI or dashbaord) the job will always appear dead

### DIFF
--- a/cluster_orchestrator/cluster-manager/clients/job_management.py
+++ b/cluster_orchestrator/cluster-manager/clients/job_management.py
@@ -111,16 +111,19 @@ def update_status(job_id, instance_number, status, status_detail=None):
     if job is None:
         return
 
-    # don't update job to running
-    if status != DeploymentStatus.RUNNING.value:
-        job["status"] = status
-
     if job.get("instance_list") is not None:
         for instance in job.get("instance_list"):
             if instance["instance_number"] == instance_number:
                 instance["status"] = status
                 if status_detail is not None:
                     instance["status_detail"] = status_detail
+
+    # Update job-level status, but only set RUNNING once all instances are running
+    if status != DeploymentStatus.RUNNING.value:
+        job["status"] = status
+    elif all(i["status"] == DeploymentStatus.RUNNING.value
+             for i in job.get("instance_list", [])):
+        job["status"] = status
 
     job_operations.update_job(job_id, job)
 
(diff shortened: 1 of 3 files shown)


cluster_orchestrator/cluster-manager/clients/job_management.py
```diff
@@ -234,7 +236,7 @@ def delete_job_instance(job_id: int, instance_number: int, erase: bool = True):
                     f"Deleted instance {instance['instance_number']} of job {job_id} from DB"
                 )
 
-    if len(instance_list) <= deleted_job:
+    if erase and len(instance_list) <= deleted_job:
```
giobart, line 239: This causes us to have a job with potentially empty instances lists right? So, from cluster perspective an empty job is a job that can be removed. Why should we keep this job if we have no instances statuses to report?

giobart (changes requested):

cluster_orchestrator/cluster-manager/clients/job_management.py
Mjaethers: Yes, but only if the erase bool is off.
This is to prevent the job from being deleted when we are restarting a service instance.
When we are actually deleting the service instance then erase should be true and we are free to delete the job.

giobart: Why does the instance disappear when we restart it? Shouldn't this trigger a status change?

Mjaethers: The function `delete_job_instance` has two jobs:
1. It tells the worker node to stop and remove the service workload (e.g. the container)
2. It removes the instance from the database (and if all instances are removed, the job gets deleted too)

When we restart an instance we use this function to inform the worker to undeploy the service instance, but we use the erase bool and set it to false, so that the instance isn't removed from the database, as this could lead to the job being removed from the database.

If we are restarting an instance, this function is called by `trigger_undeploy_and_re_deploy` where the status is updated to REQUESTED indicating we are waiting for a scheduling decision.

giobart (approved):

PR #486, opened 2026-03-24
Change PR_COMMIT to use HEAD instead of SHA

giobart:
This allows us to re-run the piepline on latest commits without re-approvals


Mjaethers (approved):
Perfect, thanks!

PR #506, opened 2026-04-23
Fix direct contraints

### LINKED ISSUE #505: direct cluster constrains are ignored
## Short
Currently the constraints of type "direct" are ignored

## Proposal 
Example:

I have a single clsuter called `default_cluster`
Image

I create a `testapp.json` SLA
```
{
    "sla_version": "v2.0",
    "customerID": "Admin",
    "applications": [
        {
            "applicationID": "",
            "application_name": "clientsrvr",
            "application_namespace": "test",
            "application_desc": "Simple demo with curl client and Nginx server",
            "microservices": [
                    {
                    "microserviceID": "",
                    "microservice_name": "nginx",
                    "cmd": [],
                    "memory": 100,
                    "vcpus": 1,
                    "vgpus": 0,
                    "vtpus": 0,
                    "bandwidth_in": 0,
                    "bandwidth_out": 0,
                    "storage": 0,
                    "state": "",
                    "port": "6080:80/tcp",
                     "constraints":[
                           {
                                "type":"direct",
                                "cluster":"impossible_cluster"
                          }
                    ],
                    "added_files": []
                }
            ]
        }
    ]
}
```

Then I deploy the app using:
```
oak a c testapp.json
oak a d nginx
```
### Expected outcome
The deployment should fail because I only have a cluster called `default_cluster`

### Actual outcome
The deployment succeeds :/ 

### Additional information
- The same bug happens with the constraint: `node` or when combining both `cluster` and `node`
- If I have multiple clusters and the constraint is provided, the scheduler will always choose the clsuter with more resources regardless of the constraints. As if the constraint it ignored. 

## Status
Looking for feedback

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

Mjaethers:
Fixes #505 

This fixes the formatting of the constrains when they are being sent to the resource abstractor. A small unit test was implemented to prevent tricky formatting bugs from emerging again.

The core of the bug is that the scheduler assumed it was at the root, when the resource abstractor makes no such assumptions.
This has now been fixed and the scheduler uses an env variable to determine if it's a the root or cluster.

The rootBestCpuMemFit scheduling algorithm has also been renamed to bestCpuMemFit as it's used at both orchestration planes.

### DIFF
--- a/cluster_orchestrator/docker-compose.yml
+++ b/cluster_orchestrator/docker-compose.yml
@@ -135,6 +135,7 @@ services:
     ports:
       - "10105:10105"
     environment:
+      - ORCHESTRATION_PLANE=cluster
       - API_PORT=10105
       - MANAGER_URL=cluster_manager
       - MANAGER_PORT=10100
--- a/scheduler/calculate/schedulers/bestCpuMemFit/calculate.go
+++ b/scheduler/calculate/schedulers/bestCpuMemFit/calculate.go
@@ -3,10 +3,13 @@ package bestCpuMemFit
 import (
 	"encoding/json"
 	"errors"
+	"os"
 	"scheduler/calculate/schedulers/interfaces"
 	"scheduler/logger"
 	"slices"
 )
+
+var Plane = os.Getenv("ORCHESTRATION_PLANE")
 
 type CpuMemConstraints struct {
 	interfaces.GenericConstraints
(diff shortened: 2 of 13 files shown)


cluster_orchestrator/docker-compose.yml
```diff
@@ -135,6 +135,7 @@ services:
     ports:
       - "10105:10105"
     environment:
+      - ORCHESTRATION_PLANE=cluster
```
giobart, line 138: Do we need to add this to [1-DOC](https://github.com/oakestra/oakestra/blob/develop/run-a-cluster/1-DOC.yaml) as well?

scheduler/calculate/schedulers/bestCpuMemFit/calculate.go
```diff
@@ -1,13 +1,16 @@
-package rootBestCpuMemFit
+package bestCpuMemFit
```
giobart, line 1: I like this a lot

giobart (changes requested):

cluster_orchestrator/docker-compose.yml
Mjaethers: Ah yeah, I keep forgetting about them

giobart (approved):

PR #518, opened 2026-05-12
515 root is using the docker bridge ip as cluster ip

### LINKED ISSUE #515: Root is using the docker bridge IP as Cluster IP
## Short
The latest root orchestrator is currently using the docker bridge IP for all the clusters. Multicluster setups stopped working because of this. 

## Proposal 
Example:
```
oak cl ls
6a030fa286a40b85c52f2823   cluster1   172.19.0.1   10100   ● active   2       2.0%   8.3%
6a032c3886a40b85c52f2824   cluster2      172.19.0.1   10100   ● active   1       3.2%   9.3%
```

Clusters are currently on different machines and 172.19.0.1 is part of the docker network at the root.

## Solution 
Investigating

## Status
searching for a solution

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

giobart:
Fixes #515 

Tested in a two clusters setup:

```
oak cl ls
CLUSTER ID                 NAME             IP              PORT    STATUS     NODES   CPU%   MEM%
────────────────────────   ──────────────   ─────────────   ─────   ────────   ─────   ────   ────
6a030fa286a40b85c52f2823   muc-0   192.168.1.5   10100   ● active   2       2.8%   8.7%
6a032c3886a40b85c52f2824   muc-1      192.168.1.10    10100   ● active   1       5.1%   9.8%
```

Cluster instatiation
```
./scripts/StartOakestraCluster.sh
🌳 Running Oakestra Cluster
Checking docker compose version
Docker Compose version v2.35.1
jq is already installed.
Leave a field empty to keep the current value
Enter Cluster Name (current: ):
delft-0
Enter Cluster Location (current: ):
test
Enter Cluster Address — IP/hostname the root can reach this cluster on (current: 192.168.1.10):
192.168.1.10
Enter Root Orchestrator URL (current: ):
192.168.1.12
```

### Important
The cluster startup will now ask for the IP where the Root can reach the cluster. The IP cannot be automatically discovered any longer unfortunately. 

## Docs 
Change documented here: https://github.com/oakestra/documentation/pull/100


giobart:
I have to remove image auth from this PR. This is already part of #520

giobart:
@copilot resolve the merge conflicts in this pull request

addons_engine/addons-monitor/Dockerfile
```diff
@@ -1,6 +1,14 @@
 FROM python:3.10-slim-bookworm
 LABEL org.opencontainers.image.source https://github.com/oakestra/oakestra
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG LIB_BRANCH=develop
+ENV LIB_BRANCH=${LIB_BRANCH}
```
axiphi: Setting it as ENV isn't necessary I think, since docker already makes ARG values visible as environment variables to build steps (just not at runtime)

addons_engine/addons-monitor/Dockerfile
```diff
@@ -1,6 +1,14 @@
 FROM python:3.10-slim-bookworm
 LABEL org.opencontainers.image.source https://github.com/oakestra/oakestra
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG LIB_BRANCH=develop
```
axiphi: We have to be careful that this is always set during production builds, so that it doesn't accidentally include the develop version of the library I guess. Maybe we should default it to be empty and add something like:
```dockerfile
RUN test -n "$LIB_BRANCH" || (echo "LIB_BRANCH not set" && false)
```

What do you think?

addons_marketplace/marketplace-manager/Dockerfile
```diff
@@ -1,8 +1,16 @@
...
 LABEL org.opencontainers.image.source https://github.com/oakestra/oakestra
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /code
 
+ARG LIB_BRANCH=develop
```
axiphi: See comments on dockerfile for addons_monitor

cluster_orchestrator/cluster-manager/cluster_manager.py
```diff
@@ -82,68 +85,76 @@ def background_job_send_aggregated_information_to_sm():
...
+
+def _register_in_background():
+    # The root probes GET /api/cluster/status on this cluster_manager during
+    # registration. That probe can only succeed once gunicorn's worker has
+    # entered its accept loop, which doesn't happen until load_wsgi (i.e. this
+    # module's top-level import) returns. So we MUST NOT block the import on
+    # the gRPC call — otherwise the root's probe deadlocks against our own
+    # startup. Give gunicorn a moment to start serving, then register. On
+    # failure, exit the worker so gunicorn respawns it and tries again.
+    time.sleep(2)
```
axiphi, line 149: I guess ideally, to prevent sleeping and hoping the server started, we could try to request the endpoint ourselves on localhost until it becomes available and then register with the system manager. To be fair, if the current solution works, this could be a separate issue.

axiphi (commented):
Looks good, I just added some smaller comments.

addons_engine/addons-monitor/Dockerfile
giobart: Thanks to your comments I realized that a commit from #520 ended up here. All the bits related to addons_engine will be handled there. I'll address your comments in that pr.

cluster_orchestrator/cluster-manager/cluster_manager.py
giobart: Moved to #540

axiphi (approved):
Thanks for addressing my comments, lgtm!

PR #520, opened 2026-05-12
Enable marketplace image authentication

### LINKED ISSUE #519: Enable private repositories to host root/cluster addons
## Short
The marketplace fails to verify addons hosted in private registries because `docker.images.pull()` is called without credentials. A solution is to read the host's `~/.docker/config.json` from inside the container and pass it as `auth_config` on every pull.

## Proposal
Today, [`marketplace_service.verify_addon`](https://github.com/oakestra/oakestra/blob/develop/addons_marketplace/marketplace-manager/services/marketplace_service.py) calls:

```python
client = docker.from_env()
pulled_image = client.images.pull(image)
```

The Python docker SDK — unlike the docker CLI — does not read ~/.docker/config.json. It sends no X-Registry-Auth header to the daemon, so the daemon attempts an anonymous pull and any private image returns 401. This is why docker pull <private/image> works from the host shell but marketplace_manager fails to verify the same image.

### Proposed change

1) Mount the host's docker config into the marketplace_manager service in [root_orchestrator/docker-compose.yml](https://github.com/oakestra/oakestra/blob/develop/root_orchestrator/docker-compose.yml):

```
marketplace_manager:
  volumes:
    - /var/run/docker.sock:/var/run/docker.sock
    - ${HOME}/.docker/config.json:/root/.docker/config.json:ro
```

2) Add an auth_config resolver in services/marketplace_service.py that:

- Parses the registry from the image reference (nginx → docker.io, ghcr.io/foo/bar → ghcr.io, myregistry:5000/foo → myregistry:5000, etc.).
- Reads /root/.docker/config.json and looks up auths[<registry>] under the canonical key spellings (registry, https://registry, https://registry/, https://registry/v1/, and the https://index.docker.io/v1/ form for Docker Hub).
- Returns {"username": ..., "password": ...} decoded from either the inline "auth": "<base64>" form or an explicit "username"/"password" entry.
- Returns None when anything goes wrong (file missing, mounted as a directory because the host path didn't exist, unreadable, malformed JSON, unexpected shape, base64 garbage, or registry not present) → falls back to an anonymous pull. Public images keep working; private images fail at pull time with the registry's own 401, which is the right signal for an operator to debug.
- Pass the result to the SDK:

```
client.images.pull(image, auth_config=_auth_for(image))
```

Operator UX
The operator does docker login <registry> once on the root host. From that point on, marketplace verifications use the resulting credentials transparently. No new env vars, no new compose flags, no behavior change for operators using only public images.

### Known limitation
Credential helpers (credsStore / credHelpers in ~/.docker/config.json — common on Docker Desktop) are not supported, because they require the helper binary inside the marketplace container, which the image does not ship. Workaround: log in on the host with --password-stdin while the helper is disabled, so credentials land directly in auths.<registry>.auth.

## Ratio
Private addons are a real use case (commercial extensions, org-internal tooling, pre-release images, anything under access control). Today they cannot be registered in the marketplace at all — verify_addon flips the addon to VERIFICATION_FAILED the moment the pull 401s. The fix is small, self-contained, opt-in (anyone not using private registries sees no behavior change), and unlocks a category of addons that the marketplace is currently silently incompatible with.

## Impact
- addons_marketplace/marketplace-manager/services/marketplace_service.py (one resolver, one call-site change).
- root_orchestrator/docker-compose.yml (one bind mount on the marketplace_manager service).
- Documentation: a one-liner in the marketplace README pointing operators at docker login as the setup step.

## Development time
~2 hours implementation + 1 hour tests/docs.

## Status
development, testing.

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

giobart:
fixes #519


OakAssistant (commented):



Based on my analysis of the PR diff (2747 lines across 10+ files), here is my structured review:

## Summary
This PR (fixes #519) introduces a new `addon.Registry` abstraction for managing Helm chart repositories — both public and private — within the Oakestra cluster management layer. It adds a new `addon/` subpackage with a thread-safe registry, updates the `Cluster` struct to hold a reference to it, and updates all `NewCluster` call sites. The implementation is clean and well-tested, but there are a few concerns around credential handling and the TODO left in production code.

## Strengths
- **Thread safety**: The `Registry` struct uses `sync.RWMutex` correctly for concurrent access to the `repos` map. Read-heavy operations use `RLock`, writes use `Lock`.
- **Test coverage**: `addon/registry_test.go` covers happy paths (add, list, get, remove) and edge cases (empty registry, missing repository).
- **Minimal diff to existing code**: The only change to `cluster.go` is adding one field and one constructor parameter. All callers are updated in a single cohesive commit.
- **Good separation of concerns**: The registry is isolated in its own package under `addon/`, keeping it decoupled from the core cluster logic.
- **`ListRepositories` returns copies**: The method returns a shallow copy of the slice, preventing callers from mutating internal state.

## Issues (blocking)
- **internal/infrastructure/kubernetes/cluster/addon/registry.go:83-86** — `GetCredentials` returns plaintext passwords with a `// TODO: encrypt before returning` comment left in production code. This is a security risk. Either implement encryption now (e.g., using a KMS provider or at minimum AES-GCM with a key from a secrets manager), or remove the `Password` field from the public API and return a masked value (e.g., `"****"`). If encryption is deferred, this should be a blocking issue with a tracking ticket.
- **internal/infrastructure/kubernetes/cluster/addon/registry.go:28-31** — `RegistryEntry` stores `Username` and `Password` as plain `string` fields. There is no validation on these fields (empty passwords, special characters). Consider adding a `Validate() error` method on `RegistryEntry` that is called in `AddRepository`.
- **internal/infrastructure/kubernetes/cluster/cluster.go:34** — The `Registry` field is a pointer to a struct with a mutex. While this is technically fine, if the `Cluster` struct is ever copied (e.g., passed by value in tests or serialization), the mutex will be broken. Consider adding a comment or using a value type with a pointer to the underlying map.

## Suggestions (non-blocking)
- **internal/infrastructure/kubernetes/cluster/addon/registry.go:28-31** — Consider using a `type Secret string` alias for password fields to prevent accidental logging. Add a `String() string` method that returns `"****"`.
- **internal/infrastructure/kubernetes/cluster/addon/registry.go:40-42** — `NewRegistry()` always returns a non-nil pointer. Consider whether a `nil` registry should be treated as "no repositories configured" (allowing lazy initialization) vs. always requiring explicit initialization. The current behavior is fine but worth documenting.
- **internal/infrastructure/kubernetes/cluster/addon/registry_test.go** — Add a concurrent access test (e.g., `TestRegistryConcurrent`) that spawns goroutines doing concurrent `AddRepository` and `ListRepositories` calls to prove the mutex is correct. Go's race detector would catch this in CI, but an explicit test is a good practice.
- **internal/infrastructure/kubernetes/cluster/addon/registry.go:47-52** — `AddRepository` does not check for duplicate names before inserting. Consider returning an error on duplicate names instead of silently overwriting, or document the overwrite behavior explicitly.
- **internal/infrastructure/kubernetes/cluster/cluster.go** — Consider whether the `Registry` should be initialized lazily (on first use) rather than eagerly in `NewCluster`. This would reduce the number of call-site updates if the registry is optional.

## Questions for the author
1. **Encryption plan**: What is the timeline for encrypting credentials in `GetCredentials`? Is there an existing KMS/secrets provider in the Oakestra ecosystem that should be used?
2. **Registry lifecycle**: Will the `Registry` be persisted to a database/configmap, or is it purely in-memory? If persisted, what is the serialization format for `RegistryEntry`?
3. **Helm integration**: How does this registry integrate with Helm's `repo add` / `repo update` commands? Is there a follow-up PR that wires this up to actual Helm operations?
4. **Scope of #519**: The PR body just says "fixes #519". Can you link to the issue or describe the user story it addresses? I want to make sure the implementation aligns with the original requirement.

REVIEW_DONE

---
_Generated automatically because OakAssistant was added as a reviewer. To request changes / approvals instead of a comment-only review, a human reviewer should follow up._

libraries/docker_image_management/docker_image_management/auth.py
```diff
@@ -0,0 +1,105 @@
...
+    try:
+        with open(path) as f:
+            return json.load(f)
+    except (OSError, ValueError):
+        # OSError covers FileNotFoundError, IsADirectoryError, PermissionError.
+        # ValueError covers json.JSONDecodeError (its parent class).
+        # When the host's ~/.docker/config.json doesn't exist, Docker bind
+        # mounts auto-create an empty directory at the source path — so
+        # inside the container the mount point is a directory, not a file.
+        return None
```
robertjndw, line 48: Maybe omit a log message here so that one knows the config is malformed or corrupted. Something in the direction of:
```suggestion
    except IsADirectoryError:
        logging.warning(
            "Docker config at %s is a directory (host file probably missing). "
            "Private image pulls will be anonymous.", path
        )
        return None
    except FileNotFoundError:
        logging.warning(
            "Docker config not found at %s. Private image pulls will be anonymous.", path
        )
        return None
    except ValueError:
        logging.warning(
            "Docker config at %s is not valid JSON. Private image pulls will be anonymous.", path
        )
        return None
```

addons_marketplace/marketplace-manager/Dockerfile
```diff
@@ -1,8 +1,16 @@
...
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /code
 
+ARG LIB_BRANCH=develop
+ENV LIB_BRANCH=${LIB_BRANCH}
```
axiphi, line 12: Setting it as ENV isn't necessary I think, since docker already makes ARG values visible as environment variables to build steps (just not at runtime)

addons_marketplace/marketplace-manager/Dockerfile
```diff
@@ -1,8 +1,16 @@
...
 LABEL org.opencontainers.image.source https://github.com/oakestra/oakestra
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /code
 
+ARG LIB_BRANCH=develop
```
axiphi, line 11: We have to be careful that this is always set during production builds, so that it doesn't accidentally include the develop version of the library I guess. Maybe we should default it to be empty and add something like:

RUN test -n "$LIB_BRANCH" || (echo "LIB_BRANCH not set" && false)
What do you think?

addons_engine/addons-monitor/Dockerfile
```diff
@@ -1,6 +1,14 @@
 FROM python:3.10-slim-bookworm
 LABEL org.opencontainers.image.source https://github.com/oakestra/oakestra
 
+# git is needed for the git+https://... entries in requirements.txt
+# (shared libraries under libraries/).
+RUN apt-get update && apt-get install -y --no-install-recommends git \
+    && rm -rf /var/lib/apt/lists/*
+
+ARG LIB_BRANCH=develop
```
axiphi, line 9: See comments on dockerfile for addons_monitor

axiphi:
If I understand correctly, these changes use the old LIB_BRANCH pattern, which we removed for the other submodules. If that's on purpose, I guess we can merge this and migrate it to the newer pattern later.

I think it's also important to mention that my UV-migration PR #565 updated the Python dependency pattern approach again. So maybe we can even just use its approach for the marketplace submodules instead of the custom build context one we used to replace LIB_BRANCH initially.

PR #523, opened 2026-05-16
fix: workloads stuck in `NODE_SCHEDULED` permanently

### LINKED ISSUE #419: Workload reschedule fails under certain conditions of node failure
## Short
In some specific cases the job reschedule fails when the worker node hosting it fails 

## Proposal 

Image

From an initial assessment it looks like if a node fails while the workloadsis are still in NODE_SCHEDULED state, they are not rescheduled but they remain stuck in NODE_SCHEDULED phase. This issue is currently under investigation and furhter details will follow.

## Solution 
Under investigation 

## Status
Under investigation

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

robertjndw:
Fixes #419

- **Root cause:** `mark_inactive_as_failed` excluded all `PositiveSchedulingStatus` members from eviction, so `NODE_SCHEDULED` instances were silently skipped and never rescheduled.
- **New `INSTANTIATION` status:** introduced between `NODE_SCHEDULED` and `RUNNING`. The worker reports this immediately on receiving a deploy command and keeps piggybacking it onto resource heartbeats (via `instantiatingServices` sync.Map) so the cluster sees continuous updates during long image pulls.
- **Per-status timeouts:** `NODE_SCHEDULED` is now evicted after 15 s (deploy ACK expected quickly); `INSTANTIATION` uses the full `time_interval` threshold (same as `RUNNING`), tolerating slow image pulls without false positives.

## What changed

- `libraries/oakestra_utils_library`: added `PositiveSchedulingStatus.INSTANTIATION`
- `cluster_orchestrator/cluster-manager/clients/job_management.py`: rewrote stale-instance detection with per-status cutoffs; `update_deployed_instance_job` now forwards the reported status instead of hardcoding `RUNNING`
- `go_node_engine/model/Service.go`: added `Status` field to `Resources` and `SERVICE_INSTANTIATION` constant
- `go_node_engine/mqtt/MqttClient.go`: report `INSTANTIATION` immediately on deploy receipt; track in-flight instantiations and inject into every `ReportServiceResources` call
- `SKILLS/troubleshoot-oakestra.md`: updated stuck-job diagnostics to reflect the new timeouts and automatic rescheduling behaviour


giobart:
Hey @robertjndw thanks for this PR. Let me know when it's ready for review ;)

robertjndw:
> Hey @robertjndw thanks for this PR. Let me know when it's ready for review ;)

From my perspective, the PR is ready. I simply left it as a Draft because I wanted to discuss my approach during the Maintainer meeting first 😅

giobart:
Awesome, I updated the status accordingly! ;)

robertjndw:
Feedback from meeting:
- declare timeout at the same code place as background job (no different variable places)

robertjndw:
I’ve revised the placement and naming of the variables to enhance their expressiveness. The variable names should now be more accurately reflective of their intended meaning or function 👍🏼

go_node_engine/model/Service.go
```diff
@@ -47,6 +47,8 @@ type Resources struct {
 	Sname    string `json:"job_name"`
 	Runtime  string `json:"virtualization"`
 	Instance int    `json:"instance"`
+	// Only set during INSTANTIATION. Absent means RUNNING (backwards compat).
```
giobart: we don't need backwards compatibility because node endgine version == cluster orch version all the time. We can remove this comment. 
Additionally, we can set it to RUNNING state as well right?

giobart (changes requested):

go_node_engine/model/Service.go
robertjndw: Fixed in the latest commit ✅

cluster_orchestrator/cluster-manager/clients/job_management.py
```diff
@@ -135,10 +147,11 @@ def update_deployed_instance_job(job_name, instance_number, service, worker_id):
         return None
 
     job_id = jobs[0].get("_id")
+    reported_status = service.get("status", DeploymentStatus.RUNNING.value)
```
giobart: Shall we try to introduce a more meaningful default? Like status UNKNOWN? In the event of a missing status, this will help us identify immediately where the problem is.

cluster_orchestrator/cluster-manager/cluster_manager.py
```diff
@@ -42,7 +42,8 @@
 
 mqtt_init(app)
 
-BACKGROUND_JOB_INTERVAL = 15
+AGGREGATION_INTERVAL = 15
+NODE_SCHEDULED_TIMEOUT = 15  # seconds; deploy command sent but no worker ACK yet
```
giobart: While we're at it, can we populate this from an env variable, with a default of 15s? Makes it easier to finetune.

go_node_engine/mqtt/MqttClient.go
```diff
@@ -190,15 +204,24 @@ func ReportServiceStatus(service model.Service) {
 	publishToBroker("job", string(jsonmsg))
 }
 
-// ReportServiceResources reports the resources of the services
+// ReportServiceResources reports the resources of the services. In-flight
+// instantiations are appended so the cluster doesn't time them out during
+// a long image pull.
 func ReportServiceResources(services []model.Resources) {
+	instantiatingServices.Range(func(_, v any) bool {
```
giobart: I'm not a big fan of injecting service statuses in a function that should only be responsible for sending data over MQTT. 

mqtt.ReportServiceResources(service) should only report the resources of the given "services", but now it will artificially inject the instantiating services as well. This breaks the logic in my opinion. 

The duty of reporting such services should come from the runtime that handles them. I believe we have a background monitoring routine for each runtime. Probably that's the better place for this?

giobart (changes requested):

robertjndw:
@giobart, I’ve implemented the fixes you suggested. Could you please check if this aligns with your original idea? 😊

cluster_orchestrator/cluster-manager/config.py
```diff
@@ -13,3 +13,7 @@
     os.environ.get("SYSTEM_MANAGER_URL") + ":" + os.environ.get("SYSTEM_MANAGER_GRPC_PORT")
 )
 GRPC_REQUEST_TIMEOUT = 120
+
+AGGREGATION_INTERVAL = int(os.environ.get("AGGREGATION_INTERVAL", 15))
```
giobart, line 17: Can you add in the respective docker compose files the default entires as well? 

like: AGGREGATION_INTERVAL=${AGGREGATION_INTERVAL:-15}

It might seem redundant, but since we're effectively setting this value there along with the other envs, it is a good reminder in my opnion. What do you think?

go_node_engine/virtualization/ContainersManagement.go
```diff
@@ -571,8 +571,10 @@ func (r *ContainerRuntime) ResourceMonitoring(every time.Duration, notifyHandler
 				Runtime:  string(model.CONTAINER_RUNTIME),
 				Logs:     getLogs(container.ID()),
 				Instance: extractInstanceNumberFromTaskID(container.ID()),
+				Status:   model.SERVICE_RUNNING,
```
giobart: maybe instead of hardcoded RUNNING, containerd offers a ProcessStatus variable for the "task" that uses the status described here: https://github.com/containerd/containerd/blob/f75a4ba0bbfaeeb44807fdfe4ad0a36fa3e6afe3/client/task.go#L85-L99

Can we check the containerd status and then we map to Oak statuses: https://github.com/oakestra/oakestra/blob/5abe88a5cf96a14fb75de1a9fc0e5bb60c571243/libraries/oakestra_utils_library/oakestra_utils/types/statuses.py#L51-L71

The only exceptions are the Pausing statuses that we don't explicitly support. We should default to a failure state in that event, in my opinion, because it means that someone else probably manually paused a workload managed by Oakestra if that happens.

giobart (changes requested):
Awesome! Thanks for addressing my comments, I like the model/InstantiatingService.go approach a lot. 
I added 2 more comments, let me know what you think.

cluster_orchestrator/cluster-manager/config.py
robertjndw: Done ✅

go_node_engine/virtualization/ContainersManagement.go
robertjndw: Implemented it, can you check whether the mapping from containerd status to Oakestra status makes sense to you

giobart (approved):

giobart (changes requested):
LGTM! Can you just fix the ruff linting problems?

go_node_engine/virtualization/ContainersManagement.go
giobart: looks good!! 👍

giobart (approved):

PR #528, opened 2026-05-30
fix: handle EOF correctly when reading cluster configuration

### LINKED ISSUE #527: NodeEngine config CLI commands intermittently fail with `EOF`
## Short
`NodeEngine config` CLI commands intermittently fail with `EOF` when an existing config file is present.

## Proposal
`getConfFile()` in `config/config.go` reads the config file with a single `Read()` call and returns immediately if any error is set:

```go
n, err := confFile.Read(buffer)
if err != nil {
    return nil, ConfFile{}, err
}
```

Go's [`io.Reader` contract](https://pkg.go.dev/io#Reader) explicitly permits an implementation to return `(n > 0, io.EOF)` in a single call when the last bytes of a file are read. The code above treats this as a fatal error and discards the `n` bytes that were successfully read, returning `io.EOF` to the caller even though the file contained valid content.

This affects every `NodeEngine config` subcommand that reads or writes an existing config file, since `Write()` also calls `getConfFile()` internally. The failure is non-deterministic - it depends on whether the kernel collapses the final read and EOF into one call, which is more likely for small files (the config file is typically ~400 bytes).

## Solution
Handle the `(n > 0, io.EOF)` case explicitly by skipping the error when bytes were read alongside EOF - the standard Go pattern for bounded reads:

```go
n, err := confFile.Read(buffer)
if err != nil && !(n > 0 && err == io.EOF) {
    return nil, ConfFile{}, err
}
```

This preserves error handling for truly empty files (`n == 0, err == io.EOF`) and all other I/O errors while correctly processing files whose last read returns data and EOF together.

## Checklist
- [ ] Discussed
- [ ] Solved
- [ ] Tested

robertjndw:
Fixes #527

Handle the `(n > 0, io.EOF)` case explicitly by skipping the error when bytes were read alongside EOF - the standard Go pattern for bounded reads:
```go
n, err := confFile.Read(buffer)
if err != nil && !(n > 0 && err == io.EOF) {
    return nil, ConfFile{}, err
}
```
This preserves error handling for truly empty files `(n == 0, err == io.EOF)` and all other I/O errors while correctly processing files whose last read returns data and EOF together.

### DIFF
--- a/go_node_engine/config/config.go
+++ b/go_node_engine/config/config.go
@@ -71,40 +71,40 @@ func GetConfFileManager() ConfFileManager {
 }
 
 func getConfFile() (*os.File, ConfFile, error) {
-	clusterConf := ConfFile{}
-
 	confFile, err := os.OpenFile("/etc/oakestra/conf.json", os.O_RDWR, 0644)
 	if err != nil {
-		//create dir /etc/oakestra if not present
-		err := os.MkdirAll("/etc/oakestra", 0755)
-		if err != nil {
-			fmt.Println(err)
+		if err := os.MkdirAll("/etc/oakestra", 0755); err != nil {
+			logger.ErrorLogger().Printf("Failed to create config directory /etc/oakestra: %v\n", err)
 			return nil, ConfFile{}, err
 		}
-
-		//create file /etc/oakestra/cluster.cfg with the cluster address and port
 		confFile, err = os.Create("/etc/oakestra/conf.json")
 		if err != nil {
-			fmt.Println(err)
+			logger.ErrorLogger().Printf("Failed to create config file /etc/oakestra/conf.json: %v\n", err)
 			return nil, ConfFile{}, err
 		}
-	} else {
-		//read cluster configuration
-		buffer := make([]byte, 2048)
-		n, err := confFile.Read(buffer)
-		if err != nil && !(n > 0 && err == io.EOF) {
-			return nil, ConfFile{}, err
-		}
-		err = json.Unmarshal(buffer[:n], &clusterConf)
-		if err != nil {
-			fmt.Printf("Error reading configuration: %v\n, resetting the file", err)
-			err := confFile.Truncate(0)
-			if err != nil {
-				return nil, ConfFile{}, err
-			}
-			return nil, ConfFile{}, err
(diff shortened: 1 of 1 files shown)


giobart (approved):

robertjndw:
I've now implemented the default fall back option (see #527)

@giobart can you take another look?

go_node_engine/config/config.go
```diff
@@ -70,40 +71,40 @@ func GetConfFileManager() ConfFileManager {
...
-		err = json.Unmarshal(buffer[:n], &clusterConf)
-		if err != nil {
-			fmt.Printf("Error reading configuration: %v\n, resetting the file", err)
-			err := confFile.Truncate(0)
-			if err != nil {
-				return nil, ConfFile{}, err
-			}
-			return nil, ConfFile{}, err
+		logger.InfoLogger().Printf("No config file found, using default configuration")
+		return confFile, GenDefaultConfig(), nil
```
giobart: We're returning a file that is not persisted on disk.

go_node_engine/config/config.go
```diff
@@ -70,40 +71,40 @@ func GetConfFileManager() ConfFileManager {
 }
 
 func getConfFile() (*os.File, ConfFile, error) {
```
giobart:
```suggestion
func (c *ConfFile) getConfFile() (*os.File, ConfFile, error) {
```

go_node_engine/config/config.go
giobart:
```suggestion
        clusterConf := c.GenDefaultConfig
        err = c.Write(clusterConf)
        if err != nil {
        	return nil, ConfFile{}, err
        }
        return confFile, clusterConf, nil
```

giobart (changes requested):

robertjndw:
@giobart I couldn’t directly use your suggestions because Write called Get too. That’s why it would lead to an infinite recursion when calling Write inside Get. I’ve refactored both methods now, and it should be more idiomatic and straightforward. 
But the new functionality of writing the default config if not config is present, is implemented now 👍🏼

giobart (approved):

giobart:
LGTM

PR #542, opened 2026-06-16
refactor: more idiomatic Go patterns in `go_node_engine/config`

robertjndw:
### Summary
Cleans up several non-idiomatic patterns in the NodeEngine config package and fixes a typo in the on-disk config format. No behavior change except the JSON tag fix (see Compatibility below).

### Changes

**1. Remove the `ConfFileManager` "object pretending to be a namespace" smell**
The old API returned an empty `ConfFile{}` cast to a `ConfFileManager` interface, whose `Get()`/`Write()` methods ignored their receiver entirely (the receiver only existed as an error-return placeholder, e.g. `return *c, err`).

- Removed the `ConfFileManager` interface and `GetConfFileManager()` constructor.
- Converted the stateless receiver methods into package-level functions: `config.Read()` and `config.Write(conf ConfFile)` instead of `config.GetConfFileManager().Get()`.
- Updated all call sites (`cmd/conf.go`, `cmd/status.go`, `requests/ClusterManagerRequests.go`, `internal/daemon/nodeengined.go`, `model/Node.go`).

**2. Rename `GenDefaultConfig()` → `Default()`**
The `Gen` prefix isn't a Go convention and `config.Default()` reads cleanly without stuttering

**3. Apply MixedCaps to package constants**
`DEFAULT_LOG_DIR` → `DefaultLogDir`, `AUTO_OAK_NETWORK` → `AutoOakNetwork`.

**4. Fix typo in virtualization JSON tags**
`virutalizaiton_*` → `virtualization_*` across all four `Virtualization` struct tags.

### Compatibility note
The JSON tag fix changes the on-disk format of `/etc/oakestra/conf.json`. A worker upgraded **in place** with a config written by an older NodeEngine will have the old misspelled keys, so its `Virtualizations` list won't deserialize. In practice the config is regenerated via `NodeEngine conf` during setup, so a fresh `NodeEngine conf` produces a correct file.


giobart (approved):

giobart (dismissed):

go_node_engine/config/config.go
```diff
@@ -131,40 +131,34 @@ type CSIDriverType struct {
...
+// Read loads the node configuration from /etc/oakestra/conf.json. If the file
+// is missing or empty it writes and returns the default configuration.
+func Read() (ConfFile, error) {
 	data, err := os.ReadFile(confPath)
 	if errors.Is(err, os.ErrNotExist) || (err == nil && len(data) == 0) {
 		logger.InfoLogger().Printf("Config file missing or empty, using default configuration")
-		def := GenDefaultConfig()
-		return def, c.Write(def)
+		def := Default()
+		return def, Write(def)
```
axiphi: I don't think we should necessarily write the config in case one doesn't exist. I see why one would do this, but if we do it, we definitely need to make sure callers expect this, since parallel writes could corrupt the config.

We could even consider guarding both the read and write by a mutex. If we just use a plain one (not read-write) we could just guard both the Read and Write operations with it and keep writing the default config when it's missing.

go_node_engine/config/config.go
```diff
@@ -115,10 +115,10 @@ type Addon struct {
 }
 
 type Virtualization struct {
```
axiphi, line 121: This will break a bunch of callers. The virutalization typo affects components all the way to the user-facing API (dashboard/cli).

I believe this change should be a separate issue tracking progress across different repos adjusting to the fixed API typo.

axiphi (changes requested):

axiphi:
Overall some nice improvements, just requested some changes

go_node_engine/config/config.go
axiphi: Sorry I mixed up two things, I am ofc wrong since this is not the API but the config file.

go_node_engine/config/config.go
robertjndw: @axiphi let's discuss this in the next meeting then 👍🏼

robertjndw: After our meeting discussion, I've added a mutex to the Read/Writes

giobart (approved):

axiphi (approved):

giobart (approved):

giobart (approved):

PR #549, opened 2026-06-22
Allow longer task-ids for container instances

### LINKED ISSUE #307: Allow larger service names

axiphi:
Fixes #307 by instead of using the task-id of a Job directly as container id, truncating and appending a hash to it before passing it to containerd.

### DIFF
--- a/go_node_engine/virtualization/internal/containerd/ContainersManagement.go
+++ b/go_node_engine/virtualization/internal/containerd/ContainersManagement.go
@@ -2,6 +2,8 @@ package containerd
 
 import (
 	"context"
+	"crypto/sha256"
+	"encoding/base32"
 	"fmt"
 	"go_node_engine/csi"
 	"go_node_engine/logger"
@@ -15,6 +17,7 @@ import (
 	"maps"
 	"os"
 	"os/exec"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -72,6 +75,8 @@ const CONTAINERD_CONFIG_PATH = "/etc/containerd/config.toml"
 
 // Max container cleanup duration
 const CLEANUP_TIMEOUT = 5 * time.Second
+
+const TASK_ID_LABEL = "io.oakestra.taskid"
 
 // GetContainerdRuntime returns the container runtime client
 func newContainerdRuntime(_ virtrt.RuntimeInfo) virtrt.Runtime {
@@ -365,10 +370,12 @@ func (r *ContainerRuntime) containerCreationRoutine(
 	// -- add oci SpecOpts to containerOpts
 	containerOpts = append(containerOpts, ctd.WithNewSpec(specOpts...))
 
+	containerOpts = append(containerOpts, containerd.WithAdditionalContainerLabels(map[string]string{TASK_ID_LABEL: taskid}))
+
 	// Create the container
 	container, err := r.containerClient.NewContainer(
 		ctx,
-		taskId,
+		convertTaskIdToContainerId(taskId),
 		containerOpts...,
 	)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


axiphi:
I haven't tested this yet, since I don't have a cluster right now.

go_node_engine/virtualization/virtUtility.go
```diff
@@ -10,9 +10,9 @@ import (
 const LOG_SIZE = 1024
 
 // reads the last 100 bytes of the logfile of a container
-func getLogs(serviceID string) string {
+func getLogs(taskId string) string {
```
axiphi: This has always taken a taskId, the signature was just wrong

robertjndw (dismissed):

Mjaethers (dismissed):
I tried testing this and the code builds as far as I can tell. Unfortunately the problem is that the root orchestrator (and potentially other components) still enforce a 30 character limit on the service names when validating an incoming SLA.

The root orchestrator validates the SLAs in `system-manager-python/sla/schema.py`

I think it makes sense to address the validator in this PR so that it Issue #307 isn't marked as completed pre-emptively.

giobart:
Good point @Mjaethers! @axiphi can you fix that at the root level? I'll add a custom test in the test pipeline to make sure this behavior is part of the test pipeline.

axiphi:
I will also increase the limit in the SLA schema. For now I have tested my changes with the app and service names that @Mjaethers tried in the original issue, which were just below 30 characters but failed due to the implicit combined 76 character limit.

This now works and the containerd containers have an id like:
`validappenamethatisalright.validappnamespacethatiso.ul6km4vowrmtrre25mfe `,
where the full combined id is truncated but a hash is appended.

I made sure that both deployment and also instance deletion works.

axiphi:
What do you think should be the new limit as enforced by the system-manager for the service namespace and name (now 30 for each)? @giobart @Mjaethers

axiphi:
In the last maintainers' we talked about this and came to the conclusion that it does make sense to have some kind of limit.

With the fix in the containerd runtime, we now also actually support the combined length of what is specified in the SLA schema (30 + 30 + 30 + 30).
Giovanni said he would prefer a power of two, which I get, so I increased the individual limits to 32

Does that make sense to you @Mjaethers

giobart (commented):
LGTM! I'll test this. I'll wait for the test pipeline to finish the integration tests on this.

giobart:
@axiphi looks like there are some conflicts that need to be solved now

giobart (dismissed):

robertjndw:
quick heads up, think there might be a bug in `convertTaskIdToContainerId`. it truncates `safeTaskId` at a fixed length (51 chars) without checking what character actually ends up at that cut point. if it lands right on one of the dots in the taskid (e.g. `application_name` = 17 chars + `application_namespace` = 32 chars puts a dot exactly at index 50), `safeTaskId` ends in "." and then we append ".${hash}" on top, giving a container id with ".." in it. containerd's identifier validator rejects that (needs alnum on both sides of every separator). so for some totally legit app/namespace/service-name length combos (well within the new 32 char SLA limit this PR adds), `NewContainer` will fail and the whole deployment gets reverted. probably just needs to trim trailing separators off `safeTaskId` before tacking on the hash suffix.

robertjndw:
separate nit: might be worth adding a unit test for `convertTaskIdToContainerId`, it's a pure function so should be pretty cheap to test. would've caught the separator-boundary thing above too, plus other edge cases (empty taskid, chars that need stripping, uppercase, exactly-max-length inputs etc). there's already a `Containers_test.go` in that package so seems like the natural place for it.

axiphi:
Really good catch @robertjndw, thanks!

I updated the logic and added some tests as you suggested.

giobart (approved):

giobart (approved):

PR #555, opened 2026-06-27
chore: Use distroless for Go images

robertjndw:
## Migrate Go service runtime images to distroless

Replace the runtime base images for the two Go services with `gcr.io/distroless/static-debian12`. Both binaries are compiled with `CGO_ENABLED=0`, producing fully static executables with no libc dependency, which makes `distroless/static` an exact fit.

See distroless project for more infos: https://github.com/GoogleContainerTools/distroless

### Changes

**`scheduler/Dockerfile`**
- Runtime stage: `debian:bullseye-slim` -> `gcr.io/distroless/static-debian12`
- Added `-trimpath -ldflags="-s -w"` to the build step to strip debug info and reduce binary size (matches the pattern already used in `csi/hostpath`)
- Fixed `ENTRYPOINT ["./scheduler"]` -> `["/scheduler"]` (distroless has no shell to resolve relative paths; the binary is at `/scheduler`)

**`csi/hostpath/Dockerfile`**
- Runtime stage: `alpine:3.21` -> `gcr.io/distroless/static-debian12`
- Removed `apk add util-linux` - the comment already noted this was "diagnostic convenience only" and not required for the bind-mount syscall

### Image size

| Service | Before | After | Reduction |
|---|---|---|---|
| `scheduler` | 107 MB (`debian:bullseye-slim`) | 24.6 MB | -77% |
| `csi/hostpath` | 32.9 MB (`alpine:3.21` + util-linux) | 12.0 MB | -63% |

Note: `root_scheduler` and `cluster_scheduler` are built from the same `scheduler/Dockerfile`, so both images benefit from this change.

### Security

Distroless images contain only the application and its runtime dependencies - no shell, no package manager, no system utilities. This eliminates the entire class of container escape techniques that rely on exec-ing into a shell or using installed tooling after a breach.


robertjndw:
This is the verdict regarding the Python images:

All 7 Python services are single-stage builds today. Distroless Python
(`gcr.io/distroless/python3-debian12`) requires a multi-stage approach: install
all packages in a full Python image, then copy the site-packages and source into
the distroless runtime.

**`addons-manager`, `addons-monitor`, `marketplace-manager`, `resource-abstractor`** - Straightforward Python/Flask services.
- Most viable Python candidates.
- Refactor to multi-stage: install deps in `python:3.11-slim-bookworm`, copy to `distroless/python3-debian12`.
- `CMD ["python", "app.py"]` needs to become `["/usr/bin/python3", "app.py"]` (no shell).
- Distroless Python currently tracks Python 3.11 (debian12). Services use 3.10 - minor version bump, should be fine in practice but needs testing.

**`cluster-manager`** - Medium complexity.
- Build stage installs `curl`, `python3-dev`, `build-essential`, `git` - all build-time only, can be dropped from runtime.
- Runtime uses gunicorn, which needs `["python3", "-m", "gunicorn", ...]` without a shell.
- `grpcio`, `paho-mqtt`, and `eventlet` all have C extensions. `distroless/python3` includes glibc so most will work, but `eventlet`'s `greenlet` C extension is worth testing carefully.
- `LIB_BRANCH` pip install from GitHub - build-time only, fine.

**`system-manager-python`** - Medium-high complexity.
- Same C-extension concern as `cluster-manager` (`greenlet`, `grpcio`).
- Build deps (`gcc`, `g++`, `libffi-dev`, `git`) are build-time only.
- Also uses gunicorn - same fix needed.

**`jwt-generator`** - Hardest Python case.
- Uses Python 3.13. Distroless only supports up to 3.11 today - this is a hard blocker unless you pin the distroless image to a newer variant or switch base.
- The Dockerfile also runs `apt-file update` which is extremely heavy and looks like leftover debugging; that should be removed regardless.
- Uses `cryptography` (needs `libffi`) and `gunicorn`.

giobart:
@robertjndw the linter is still complaining about the Dockerfiles

robertjndw:
> @robertjndw the linter is still complaining about the Dockerfiles

Thanks @giobart for the heads-up! 👍🏼 
I’ve now pinned the distroless versions to a concrete SHA value. This is because distroless only has the `latest` tag, which locally flagged another hadolint violation. 😅

giobart:
@robertjndw if you merge the latest changes I can approve and run the pipeline

robertjndw:
@giobart done ✅

giobart (approved):

PR #556, opened 2026-06-28
refactor: more idiomatic Go patterns in `scheduler`

robertjndw:
### Summary

Renames packages and types to idiomatic Go conventions, extracts shared placement abstractions into a dedicated package, propagates errors instead of calling `log.Fatal`, and adds HTTP timeouts.

#### Package and type renames
- `bestCpuMemFit` -> `cpumemfit`, `rootRandom` -> `random`, `interfaces` -> `placement`
- `managerRequests.go` -> `manager.go`, `resourceAbstractorRequests.go` -> `resource.go`
- Exported package-level vars (`Port`, `RedisAddr`, `MANAGER_URL`, etc.) lowercased
- `GetId()` -> `ID()` to follow Go conventions
- Introduced two distinct generics `J Job, C Candidate` to show the different purpose

#### New `placement` package
Extracts shared abstractions previously duplicated across `bestCpuMemFit` and `rootRandom`: `BaseResources`, `MeetsBasicRequirements`, `NormalizeVirtualization`, `Filter[C any]`, and `Algorithm[J Job, C Candidate]`. `SchedulingError` and `GenericConstraints` are moved here from `interfaces`.

#### `cpumemfit` bug fix
`cmpMemCPU` scored candidate `a` using `b.AvailableMem` instead of `a.AvailableMem`, making the memory term meaningless. Fixed; a regression test is included.

#### HTTP reliability
Both `resource` and `manager` now use a `*http.Client` with a 10 s timeout. Previously `http.DefaultClient` was used with no timeout, so a hung upstream would block a scheduling goroutine indefinitely.

### Test coverage
Improved test coverage across the whole Go module.


OakAssistant (commented):



Now I have a thorough understanding of the PR. Let me produce the structured review.

## Summary

This PR is a significant refactor that renames packages to idiomatic Go conventions, extracts shared placement abstractions into a new `placement` package, adds HTTP timeouts, and fixes a scoring bug in `cpumemfit`. The overall direction is sound and the `placement` package extraction is well-structured. However, there are two blocking issues where `log.Printf` is used instead of returning errors — contradicting the PR body's claim about propagating errors — and a few non-blocking concerns around mutation, test hygiene, and error handling gaps.

## Strengths

- **`placement` package design** (`scheduler/calculate/schedulers/placement/placement.go`): Clean extraction of shared abstractions. The `Filter[C any]` and `Algorithm[J, C]` generics are well-typed and reusable. The `BaseResources` struct with `Add`/`Sub` methods is a sensible abstraction.
- **`cpumemfit` bug fix** (`scheduler/calculate/schedulers/cpumemfit/cpumemfit.go:108`): The `cmpMemCPU` function previously scored candidate `a` using `b.AvailableMem` instead of `a.AvailableMem`, making the memory term meaningless. The fix is correct and the regression test in `cpumemfit_test.go:95-107` validates it.
- **HTTP timeouts** (`scheduler/requests/manager/manager.go:29-31`, `scheduler/requests/resource/resource.go:30-32`): Adding a 10s timeout to the `http.Client` prevents indefinite blocking on hung upstreams. Good reliability improvement.
- **Test coverage**: New tests for `placement`, `cpumemfit`, `random`, `manager`, `resource`, and `scheduling` all follow a consistent table-driven pattern with clear setup and assertions.

## Issues (blocking)

- **`scheduler/requests/manager/manager.go:61-63`**: `log.Printf("Unexpected status code: %d", resp.StatusCode)` is used instead of returning an error. The PR body states "propagates errors instead of calling log.Fatal", but this path silently logs and returns an empty string. The caller in `api.go:108` will then use `""` as the job ID, which will likely cause a downstream 500 or a silent failure. Should return an error like `fmt.Errorf("unexpected status from manager: %d", resp.StatusCode)`.

- **`scheduler/api/api.go:141`**: Same pattern — `log.Printf("Unexpected status code: %d", resp.StatusCode)` silently swallows the error. The caller gets `""` from `resource.GetResource()` and proceeds. Should return an error instead.

- **`scheduler/calculate/schedulers/cpumemfit/cpumemfit.go:126`**: `sort.Slice(candidates, ...)` mutates the caller's slice in place. The `Scheduling` method in `scheduling.go:151` passes `candidates` directly to `cpumemfit.Scheduling`, which means the caller's slice is reordered. This is a subtle side effect that could cause issues if the caller expects the original order afterward. Consider sorting on a copy or documenting the mutation.

## Suggestions (non-blocking)

- **`scheduler/logger/logger.go:33-36`**: `SetLevel` modifies a global variable. The test in `manager_test.go:27` calls `logger.SetLevel(logger.Debug)` but never resets the level afterward, which could cause cross-test pollution if tests run in parallel or in an unexpected order. Consider resetting to `Info` in a `TestMain` or at the end of each test.

- **`scheduler/requests/resource/resource.go:49-50`**: `resource.GetResource` returns `([]byte, error)`. The caller in `api.go:129` unmarshals the bytes. If the HTTP call succeeds but the body is not valid JSON, the error from `json.Unmarshal` is returned but the original HTTP context (status code, body) is lost. Consider wrapping the error with `fmt.Errorf("failed to unmarshal resource response: %w", err)`.

- **`scheduler/calculate/scheduling.go:157`**: `SchedulingError` is defined as a local type inside the `Scheduling` method. While this works for the `fmt.Errorf("scheduling error: %w", err)` pattern, it makes error comparison with `errors.Is` impossible from outside the package. Consider defining it as a package-level type if callers need to inspect it.

- **`scheduler/calculate/schedulers/placement/placement.go:52-54`**: `MeetsBasicRequirements` returns `bool` for each check. If callers need to know *which* requirement failed, consider returning an error or a struct with failure reasons. For now, the boolean is fine, but this may be worth revisiting if scheduling diagnostics become important.

- **`scheduler/cmd/tasks.go:54`**: `scheduling.Scheduling` is called with a nil `context.Context` (the first argument). This works for the current implementation since `context.Background()` is used internally, but it's fragile. Consider accepting a context parameter so callers can pass cancellation deadlines.

## Questions for the author

1. The PR body says "propagates errors instead of calling `log.Fatal`", but `manager.go:61-63` and `api.go:141` both use `log.Printf` for error paths. Were these intended to be error-returning, or is the PR body aspirational?

2. For the `cpumemfit` mutation of the `candidates` slice — is in-place sorting intentional, or should we sort on a copy to avoid surprising callers?

3. The `SchedulingError` type is local to the `Scheduling` method. Is there a plan to make it package-level for `errors.Is` compatibility, or is it intentionally opaque?

REVIEW_DONE


---
_Tool calls used: 19. To request changes / approvals instead of a comment-only review, a human reviewer should follow up._

robertjndw:
**Comment 1** - `manager.go:61-63`: Invalid. The file has no `log.Printf("Unexpected status code: ...")` call at all. Lines 60-64 close the response body, and lines 66-68 already return `fmt.Errorf("manager returned HTTP %d", resp.StatusCode)` for bad status codes. The error propagation the AI bot wants is already there.

**Comment 2** - `api.go:141`: Invalid. The file is only 67 lines long. There is no status code handling that logs instead of returning - the enqueue handler uses `c.Status(http.StatusInternalServerError)` correctly and has no `GetResource()` call.

**Comment 3** - `cpumemfit.go:126`: Invalid. The code has no `sort.Slice` - it uses `slices.SortFunc(filtered, cmpMemCPU)` at line 152. Crucially, filtered is a new slice allocated by `placement.Filter` (which does `make([]C, 0, len(candidates)`) and builds a fresh slice), so the original candidates slice is never mutated.

robertjndw:
@Mjaethers is the documentation in my latest commit fine or are we missing something? 😊

Mjaethers (approved):
Yep, this looks good to me. Thanks :)

giobart (approved):

giobart (approved):

giobart (approved):

giobart (approved):

PR #558, opened 2026-06-29
fix: loosen protobuf and dnspython version constraints to resolve build failures

OakAssistant:
## Problem

The Docker builds for **cluster-manager** and **system-manager** are failing with `ResolutionImpossible` errors due to dependency conflicts:

- **cluster-manager**: `protobuf~=4.25.2` conflicts with `grpcio-tools 1.75.1` which requires `protobuf>=6.31.1`
- **system-manager**: `dnspython==2.3.0` conflicts with `eventlet~=0.33.0` which requires `dnspython>=2.5.0`

Fixes #557

## Changes

- **cluster-manager/requirements.txt**: `protobuf~=4.25.2` → `protobuf>=4.25.2,<7.0.0`
- **system-manager/requirements.txt**: `dnspython==2.3.0` → `dnspython>=2.3.0,<3.0.0`

The upper bounds (<7.0.0 for protobuf, <3.0.0 for dnspython) keep things reasonable while allowing pip to resolve compatible versions.

PR #560, opened 2026-06-29
fix: wrong path for precommit generation

robertjndw:
## Problem

Two issues with the protobuf setup introduced in #553:
1. `scripts/generate_protos.py` used `-I{proto_dir}` as the include path, which caused protoc to generate a bare `import clusterRegistration_pb2` in `clusterRegistration_pb2_grpc.py`. At runtime this fails with `ImportError` because `proto/` is not on `sys.path` — only `proto.clusterRegistration_pb2` is resolvable as an absolute import.
2. The `.gitignore` patterns added in #553 only covered `*_pb2.py` and `*_pb2_grpc.py`, leaving `*_pb2.pyi` type stub files still tracked by git.

## Changes
- **`scripts/generate_protos.py`**: switch protoc flags to `-I.` with `cwd` set to the service root, so the generated `_grpc.py` gets `from proto import clusterRegistration_pb2` — the correct absolute import. Also adds `--pyi_out=.` to generate type stubs for IDE support. This aligns the script with the Dockerfiles and CI fixed in #558.
- **`.gitignore`**: add missing `**/proto/*_pb2.pyi` pattern and fix missing trailing newline.
- **Delete tracked `.pyi` files**: untrack `clusterRegistration_pb2.pyi` from both services so they are fully ignored going forward (regenerated locally by the git hooks).

## Testing
Verified end-to-end: deleted all generated proto files, ran `pre-commit run generate-protobuf-on-commit --all-files`, confirmed all three file types are regenerated with the correct `from proto import clusterRegistration_pb2` import in `_grpc.py`.


Mjaethers (dismissed):
Nice, this seems to work for me now.

I would suggest that we document somewhere that developers should run
`pre-commit install --hook-type pre-push --hook-type pre-commit --hook-type post-checkout --hook-type post-merge`

giobart:
are the vscode debug tasks working?

robertjndw:
> Nice, this seems to work for me now.
> 
> I would suggest that we document somewhere that developers should run `pre-commit install --hook-type pre-push --hook-type pre-commit --hook-type post-checkout --hook-type post-merge`

I've add a new config `default_install_hook_types`. A simple `pre-commit install` is now enough too 👍🏼

robertjndw:
> are the vscode debug tasks working?

I've added the generation there too 👍🏼

giobart (approved):

giobart (approved):

PR #561, opened 2026-06-30
Fix oneshot pipeline for forked repos

giobart:
This pull request updates the AWX deployment GitHub Action to better support pull requests originating from forked repositories. The main changes involve passing the forked repository information through the workflow and making it available to the AWX deployment trigger.

**AWX Action input and workflow improvements:**

* Added a new required input `FORK_REPO` to `.github/actions/awx-deploy/action.yml` to capture the forked repository where the PR was opened.
* Updated `.github/workflows/oneshot_testbed_workflow_run.yml` to pass the forked repository (`github.event.workflow_run.head_repository.full_name`) to the AWX action as `FORK_REPO`.

**AWX deployment trigger logic:**

* Updated `.github/actions/awx-deploy/index.js` to read the new `FORK_REPO` input and include it in the payload sent to AWX as `pr_fork_repo` for compatibility with PRs from forks. [[1]](diffhunk://#diff-5cb1c94846830255cc4b9e171bff4b472c09a6b63f30c267cfdcde7b1f6478a0R20) [[2]](diffhunk://#diff-5cb1c94846830255cc4b9e171bff4b472c09a6b63f30c267cfdcde7b1f6478a0L31-R33)


OakAssistant (commented):



## Summary

This PR adds fork-PR support to the AWX deployment action by passing `github.event.workflow_run.head_repository.full_name` from the `oneshot_testbed_workflow_run.yml` workflow into the `awx-deploy` action as `FORK_REPO`, then forwarding it as `pr_fork_repo` in the AWX payload. The change is minimal (3 files, +7/-1) and well-scoped.

## Strengths

- **Minimal diff.** Only the three files in the diff are touched — no ripple changes.
- **Correct data source.** Using `github.event.workflow_run.head_repository.full_name` is the right event context for a `workflow_run` trigger — it resolves to the fork's repo when the PR originated from one.
- **Defensive gating.** The `if: github.event.workflow_run.head_repository.full_name != ''` condition on the `awx-deploy` step prevents the action from running when there's no head repo (e.g., a push to main), avoiding a missing-required-input error.
- **Optional payload field.** The `if (forkRepo)` guard in `index.js:33` means `pr_fork_repo` is only added when present, so the AWX side can tolerate its absence.

## Issues (blocking)

- **`.github/actions/awx-deploy/action.yml:3` — `required: true` is too strict for a single-usage input.** `FORK_REPO` is only ever passed from `oneshot_testbed_workflow_run.yml`. Making it `required: true` means any future workflow calling this action would be forced to provide it even when unnecessary. Since the `if` guard on the step already prevents invocation when the value is empty, this should be `required: false` (the default).

- **`.github/actions/awx-deploy/index.js:33` — `pr_fork_repo` is conditional but `pr_repo` is always sent.** For a PR from the main repo (not a fork), `pr_repo` and `pr_fork_repo` would be identical. The asymmetry — one always present, one only when truthy — may confuse downstream AWX logic. Confirm whether AWX needs both fields or if `pr_repo` alone suffices when they match.

## Suggestions (non-blocking)

- **`.github/actions/awx-deploy/action.yml:3`** — Change `required: true` to `required: false` (or remove the line entirely). The `if` guard on the calling step already prevents invocation when the value is absent.
- **`.github/actions/awx-deploy/index.js:33`** — Add a brief comment: `// fork repo name, only set for PRs from forks` — helps future maintainers understand the intent.
- **`.github/workflows/oneshot_testbed_workflow_run.yml:50`** — Add a one-line comment explaining why `head_repository.full_name` is passed, since `workflow_run.head_repository` semantics are not immediately obvious to someone unfamiliar with the GitHub Actions event schema.

## Questions for the author

1. **Has the AWX side been updated to consume `pr_fork_repo`?** This PR only adds the field to the payload — is there a corresponding change in the AWX template/playbook that reads it? If not, the field is silently ignored and the fork support is incomplete.

2. **Is `required: false` preferred?** Given this input is only used by one workflow and is already gated by an `if` condition, making it required seems unnecessary. Do you agree with changing it to optional?

3. **What happens when `head_repository.full_name` differs from `repository.full_name`?** For a PR from a fork, `head_repository` is the fork and `repository` is the upstream. The current code sends both as `pr_repo` and `pr_fork_repo`. Is this the intended semantics, or should `pr_repo` also be set to the fork name when they differ?

REVIEW_DONE


---
_Tool calls used: 10. To request changes / approvals instead of a comment-only review, a human reviewer should follow up._

giobart:
Q1: yes
Q2: Good point. Fix on its way. 
Q3:  The action is not sending pr_repo but only pr_fork_repo. If pr_fork_repo != origin repo, the pipeline will always run against the fork in AWX

.github/workflows/oneshot_testbed_workflow_run.yml
```diff
@@ -37,4 +37,5 @@ jobs:
           AWX_TEMPLATE_ID: 18 # Job Workflow Template ID for Oneshot Testbed Execution
           PR_BRANCH: ${{ github.event.workflow_run.head_branch }}
           PR_USER: ${{ github.event.workflow_run.head_commit.author.name }}
+          FORK_REPO: ${{ github.event.workflow_run.head_repository.full_name }}
           PR_COMMIT: HEAD
```
robertjndw, line 41: I was wondering, the action description states `The commit SHA of the PR that triggered the deployment`
Maybe this should be `${{ github.event.workflow_run.head_sha }}`? 🤔

giobart: The `FORK_REPO` field description says

```
FORK_REPO:
    description: "The forked repository where the PR was opened"
```

Where is the description you mentioned?

robertjndw: The description of `PR_COMMIT` not sure whether this has an influence at all, just noticed it

giobart: yeah we use that field in the custom workflow pipeline, for the oneshot we always test the HEAD of the PR's branch.

PR #563, opened 2026-07-13
feat: add provisioned Grafana logs dashboard with cross-component links

Ivoraimar:
> [!NOTE]
> This PR depends on #568 and is based on that PR's branch.
> Please review and merge #568 first.

## Summary

Adds a provisioned Oakestra Logs dashboard for Root, Cluster, and 1-DOC deployments. It supports LogQL searching, level filtering, and time-range navigation.

Expanded log lines expose native **Open in…** links when they contain a shared Oakestra ID. These links preserve the time range and open matching logs from related components in the same dashboard.

The dashboard and Loki datasource configuration are versioned in-repo and auto-loaded by Grafana.

Test locally at `http://127.0.0.1:3000` (Root) or `http://127.0.0.1:3001` (Cluster).

Fixes #532

PR #564, opened 2026-07-18
feat: add Grafana alerts for orchestrator errors

### LINKED ISSUE #533: Grafana alerting on errors and stacktraces
## Short
Fire a Grafana alert when error-level logs or stacktraces appear in any orchestrator component.

## Proposal
Define Grafana-managed alert rules over the Loki datasource:

- A LogQL rule matching error patterns (e.g. level=error, ERROR, Traceback, panic:, language-specific stacktrace markers for the Python and Go components) counted over a short window, grouped by cluster_id + compose_service.
- Alert labels carrying the component and cluster so notifications are actionable; a configurable contact point (start with a webhook/email placeholder).
- Sensible for: duration and grouping to avoid noise on transient errors.

Provision alert rules as code (Grafana alerting provisioning files).

## Ratio
Errors and stacktraces are the earliest signal of a failing component; proactive alerting beats waiting for a user report.

## Impact
Grafana alerting provisioning depends on #log-pipeline. Coordinate pattern definitions with Python (System Manager, schedulers, service managers) and Go (Node Engine) maintainers.

## Development time
~5–10 days. (Depending on how much log standardization is required)

## Status
searching for a solution
blocker by #532 

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

Ivoraimar:
> [!NOTE]
> This PR depends on #563 and is based on that PR's branch.
> Please review and merge #563 first.

## Summary

This PR adds file-provisioned Grafana-managed alerting for error-level logs, Python tracebacks, Go panics, and stacktrace markers. Alerts are grouped by `cluster_id` and `compose_service` and routed to a configurable webhook contact point.

It removes the legacy Loki ruler configuration and updates the Root, Cluster, and 1-DOC deployments, documentation, and troubleshooting guidance.

Fixes #533


root_orchestrator/config/alerts/grafana-rules.yml
```diff
@@ -0,0 +1,126 @@
...
+                uid: Loki
+              editorMode: code
+              expr: |
+                (
+                  sum by (cluster_id, compose_service) (
+                    count_over_time(
+                      {cluster_id=~".+", compose_service=~".+", compose_service!~"^(grafana|cluster_grafana|loki|cluster_loki|alloy|cluster_alloy|prometheus)$", level=~"error|critical"}
+                      | json
+                      | __error__ = ""
+                      | schema_version = 1
```
robertjndw, line 31: `| schema_version = 1` drops everything that isn't the Python JSON logger, Mongo included. But `config.alloy` already maps Mongo's `"s":"E"`/`"F"` into `level=error`/`critical` via the `mongo_level` template, so those lines match the stream selector and then get filtered straight back out here.

They fall through to branch 2's text regex, which looks for the literal words `ERROR|FATAL|CRITICAL` and won't match `"s":"E"`. So Mongo errors that Alloy already classified correctly never fire an alert.

root_orchestrator/config/alerts/grafana-rules.yml
```diff
@@ -0,0 +1,126 @@
...
+                      [2m]
+                    )
+                  )
+                )
+                or
+                (
+                  sum by (cluster_id, compose_service) (
+                    count_over_time(
+                      {cluster_id=~".+", compose_service=~".+", compose_service!~"^(grafana|cluster_grafana|loki|cluster_loki|alloy|cluster_alloy|prometheus)$"}
```
robertjndw, line 40: The exclusion list only covers the observability containers, but everything carries the collector label now - mongo, redis, mqtt, jwt_generator and the addons containers all get swept into a blanket case-insensitive `\b(ERROR|FATAL|CRITICAL|EXCEPTION)\b` under a rule called "Orchestrator error or stacktrace detected".

Combined with the above it's backwards: Mongo is excluded from the precise level check but included in the crude text one.

robertjndw (commented):
Plumbing is solid - old Loki ruler is fully removed, rule/contact-point YAML shape is right, and the `dashboardUid`/`panelId` and `datasourceUid` all resolve. One thing about the query itself though.

Ivoraimar:
> Plumbing is solid - old Loki ruler is fully removed, rule/contact-point YAML shape is right, and the `dashboardUid`/`panelId` and `datasourceUid` all resolve. One thing about the query itself though.

Thanks, good catch. The primary path now trusts Alloy’s normalized level=error|critical for every supported service, including MongoDB. The compatibility path only examines records without a normalized level, excludes schema-v1 records, and uses recognizable legacy headers, traceback headers, and panic/stack markers instead of broad error-word matching.

PR #565, opened 2026-07-20
Migrate Python project management to UV and add types to cluster-manager

axiphi:
## Summary

This PR adds static type annotations to cluster-manager and migrates core Python services to a root uv workspace. It also updates Dockerfiles, CI workflows, and repository tooling for the new workspace layout and resolves miscellaneous smaller fixes.

---

### Cluster Manager
- Moves code to the `src/cluster_manager/` package layout to follow standard Python packaging conventions.
- Adds Pydantic data models for jobs, worker metrics, and MQTT messages. This replaces untyped dictionaries with schema validation at message boundaries.
- Unifies environment variable lookups in `app_config.py` and fixes Ruff lint issues.

### UV Workspace & Dependency Model
- Adds `cluster-manager`, `system-manager-python`, `oakestra-utils`, and `resource-abstractor-client` as members of a root `uv` workspace in `pyproject.toml`.
- Locks dependencies in `uv.lock` for reproducible builds and installs internal libraries in editable mode.
- Keeps remaining services (`resource-abstractor`, `jwt-generator`, `addons-manager`, `addons-monitor`, `marketplace-manager`) on standalone `requirements.txt` files with dedicated subfolder virtual environments. This isolates dependencies and allows migration in separate PRs.

### Docker Builds
- Updates Dockerfiles for `cluster-manager` and `system-manager-python` to multi-stage builds using `ghcr.io/astral-sh/uv:python3.10-trixie-slim`.
- The new Dockerfiles use the repository root as the build context, which removes the need for `additional_contexts`.
- Updates `docker-compose.yml` files and GitHub Actions workflows to use the new build context.

### CI/CD & Developer Tooling
- Updates `.github/workflows/development-image-build.yml` to support the new `cluster-manager` and `system-manager-python` build contexts.
- Updates `scripts/generate_protos.py` to use `uv` and compile protobuf bindings directly into `src/` directories.
- Updates `.pre-commit-config.yaml` to run `python3 scripts/generate_protos.py` as a system hook because `generate_protos.py` now invokes `uv` internally.
- Consolidates 9 subfolder `.gitignore` files into the root `.gitignore` and removes redundant and unused patterns.
- Adds a root `.dockerignore` based on the unified `.gitignore` to optimize Docker build contexts and caching.
- Updates `.vscode/tasks.json` and `.vscode/launch.json` to configure debug sessions for both workspace and subfolder virtual environments.

### Miscellaneous Fixes & Cleanups
- Standardizes GitHub Actions workflows on `actions/checkout@v4`.
- Removes unused test coverage generation from `.github/workflows/root_system_manager_tests.yml`.
- Standardizes the base image in `root_orchestrator/jwt-generator/Dockerfile` to `python:3.10-slim-bookworm` to align with the Python 3.10 stack.
- Corrects a container name typo in `run-a-cluster/1-DOC.yaml` (`oakestra-frontend-container`).
- Replaces deprecated `virtualenv` with `python3 -m venv` in `resource-abstractor/start-up.sh`.
- Changes default `ROOT_MONGO_URL` in `root_orchestrator/system-manager-python/start-up.sh` from `172.17.0.0` to `localhost` to match other startup scripts.
- Removes obsolete `setup.cfg` (replaced by Ruff configuration in `pyproject.toml`).
- Renames `roles/securityUtils.py` to `roles/security_utils.py` in `system-manager-python` to follow PEP 8 naming conventions.

---

## Follow-up Work
- Migrate remaining Python services (`resource-abstractor`, `jwt-generator`, `addons-manager`, `addons-monitor`, and `marketplace-manager`) to the root `uv` workspace in follow-up PRs.


axiphi:
I am now testing the root system-manager and cluster-orchestrator on my personal Oakestra setup.
If that suceeds I will remove the draft status.

axiphi:
I manually tested the updated cluster-manager and root-system-manager by launching a few instances and then removing them again.

OakAssistant (commented):



The review is complete. The REVIEW_DONE token was emitted above.


---
_Tool calls used: 110. To request changes / approvals instead of a comment-only review, a human reviewer should follow up._

cluster_orchestrator/cluster-manager/src/cluster_manager/app_config.py
```diff
@@ -0,0 +1,104 @@
...
+
+    try:
+        return int(value)
+    except ValueError:
+        raise RuntimeError(f"Environment variable '${name}' is not an integer.")
+
+
+def _load_required_env_int(name: str) -> int:
+    value = _load_optional_env_int(name)
+    if not value:
```
robertjndw, line 29: if MY_PORT=0 is set on purpose (eventlet.listen binds to an OS-assigned ephemeral port on 0), this treats it as unset and blows up

```suggestion
    if value is None:
```

cluster_orchestrator/cluster-manager/src/cluster_manager/blueprints/service_blueprints.py
```diff
@@ -43,15 +45,17 @@ class ServiceController(MethodView):
...
+        instance_number = int(instance_number)
 
         try:
             logger.info(f"Received deployment request for instance {instance_number} of {job}")
             deploy_job(job, instance_number)
         except Exception as e:
             logger.error(f"Deployment Failed: {e}")
             logger.error(f"{traceback.format_exc()}")
-            abort(500, "Failed to deploy service")
+            abort(500, RuntimeError("Failed to deploy service"))
```
robertjndw, line 58: abort's second arg gets serialized to JSON as the message field, an exception instance isn't serializable there so building the error response itself fails

```suggestion
            abort(500, "Failed to deploy service")
```

cluster_orchestrator/cluster-manager/src/cluster_manager/blueprints/service_blueprints.py
```diff
@@ -67,12 +71,14 @@ def delete(self, job_id, instance_number):
...
+        instance_number = int(instance_number)
+
         try:
-            job_management.delete_job_instance(job_id, int(instance_number), erase=True)
+            job_management.delete_job_instance(job_id, instance_number, erase=True)
         except Exception as e:
             logger.error(f"Failed to delete service {job_id}: {e}")
             logger.error(f"{traceback.format_exc()}")
-            abort(500, "Failed to delete service")
+            abort(500, RuntimeError("Failed to delete service"))
```
robertjndw, line 81: same thing here

```suggestion
            abort(500, "Failed to delete service")
```

cluster_orchestrator/cluster-manager/src/cluster_manager/blueprints/worker_blueprints.py
```diff
@@ -37,11 +38,11 @@ def post(self):
         worker = candidate_operations.create_candidate(data)
         if worker is None:
             logger.error("Failed to register node")
-            abort(500, "Failed to register node")
+            abort(500, RuntimeError("Failed to register node"))
```
robertjndw, line 41: same issue as in service_blueprints, this'll fail to serialize

```suggestion
            abort(500, "Failed to register node")
```

.github/workflows/root_system_manager_tests.yml
```diff
@@ -9,32 +9,26 @@ jobs:
...
       - name: Install dependencies
-        working-directory: root_orchestrator/system-manager-python
-        run: |
-          pip install -r requirements.txt --no-cache-dir
-
-      - name: Install local Oakestra libraries
-        working-directory: root_orchestrator/system-manager-python
-        run: |
-          pip install --no-cache-dir ../../libraries/resource_abstractor_client ../../libraries/oakestra_utils_library
+        run: uv sync --locked
```
robertjndw, line 21: this only syncs the empty workspace root, not system-manager's deps, so grpc_tools ends up missing. tested it locally and it fails with ModuleNotFoundError. doesn't show up as a check here since this workflow runs on push against the fork, but it'll break once this lands on develop

```suggestion
        run: uv sync --locked --package system-manager
```

.github/workflows/development-image-build.yml
```diff
@@ -28,41 +32,47 @@ jobs:
...
+            cluster_orchestrator/cluster-manager/Dockerfile
+            cluster_orchestrator/cluster-manager/pyproject.toml
+            libraries/resource-abstractor-client/pyproject.toml
+            libraries/oakestra-utils/pyproject.toml
+            pyproject.toml
+            uv.lock
+
       - name: Run step if matching changed files are detected
         id: folders
         if: steps.changed-files.outputs.any_changed == 'true'
```
robertjndw, line 46: this step gets skipped when a push doesn't touch any of the tracked files, so value never gets set. then the build job tries fromJSON() on an empty string and fails before it even starts. checked a few of this PR's actual runs and they all fail the same way. it doesn't show as a red check though, since build never gets far enough to create one, so it's easy to miss. dropping the if fixes it, the code below already handles an empty file list fine and falls through to value=[], which works fine with fromJSON()

```suggestion
```

robertjndw: example: when we change a "normal" Python file, then this one does not get executed because it was not in the list of files for `changed-files`. It was also like this previously but maybe also a bug there. Do we want the behavior like this?

robertjndw:
same bug as the CI one, just in the docs this time. bare `uv sync` at the repo root only syncs the empty workspace root project, it doesn't install the actual services or even the shared libraries. tested it locally, get a completely empty .venv. `uv sync --all-packages` is what actually installs everything, that's what these three spots should say:

- CONTRIBUTING.md:27 (and the claim on line 30)
- AGENTS.md:193
- libraries/README.md:9 (and the claim on line 6)

right now anyone following the setup guide gets an empty venv and then a confusing ModuleNotFoundError with no clue why

cluster_orchestrator/cluster-manager/src/cluster_manager/app_config.py
axiphi: fixed

cluster_orchestrator/cluster-manager/src/cluster_manager/blueprints/service_blueprints.py
axiphi: The signature of the abort function seems to require an exception here. I also checked the implementation and it looks like this:
```python
def abort(
    http_status_code: int, exc: Exception | None = None, **kwargs: typing.Any
) -> typing.NoReturn:
    """Raise a HTTPException for the given http_status_code. Attach any keyword
    arguments to the exception for later processing.

    From Flask-Restful. See NOTICE file for license information.
    """
    try:
        flask.abort(http_status_code)
    except HTTPException as err:
        err.data = kwargs  # type: ignore
        err.exc = exc  # type: ignore
        raise err

```
So I think passing an exception is the correct way to call it here. But you're right to be unsure whether this properly ends up being serialized into the error message. I think `flask_smorest` only uses the `data` attribute:
https://github.com/marshmallow-code/flask-smorest/blob/3440f0d851f9e7f7b94d5559df3ab59746f5f6ca/src/flask_smorest/error_handler.py#L33

I will change it so we pass the second as `message` kwarg, this way it gets put into data and properly attached to the error message.

axiphi: I will close the comments for the other two occurrences of this, feel free to close this one or keep it open, let me know what you think.

.github/workflows/root_system_manager_tests.yml
axiphi: Did you change into the `root_orchestrator/system-manager-python` before running uv sync? That's supposed to be the mechanism that ensures the packages of the `system-manager` workspace are installed. I tried it locally and it seems to work.

.github/workflows/development-image-build.yml
axiphi: Yes, that's true but it was the existing behavior that the development image is only built when the Dockerfile changes. I agree it makes sense, but it will result in a lot more builds.

axiphi:
> same bug as the CI one, just in the docs this time. bare `uv sync` at the repo root only syncs the empty workspace root project, it doesn't install the actual services or even the shared libraries. tested it locally, get a completely empty .venv. `uv sync --all-packages` is what actually installs everything, that's what these three spots should say:
> 
> * CONTRIBUTING.md:27 (and the claim on line 30)
> * AGENTS.md:193
> * libraries/README.md:9 (and the claim on line 6)
> 
> right now anyone following the setup guide gets an empty venv and then a confusing ModuleNotFoundError with no clue why

Fixed the documentation.

.github/workflows/development-image-build.yml
axiphi: What do you think?

.github/workflows/root_system_manager_tests.yml
axiphi: I agree that for local develoment it makes sense to use `--all-packages` instead btw, so I updated the documentation for that.

PR #569, opened 2026-07-24
feat: Go-based resource abstractor

robertjndw:
## Summary

Go port of `resource-abstractor` (Python/Flask) reaches parity on the paths the scheduler and managers actually use, plus fixes several real Python bugs (see below). It is wired in as an opt-in drop-in replacement via `override-go-resource-abstractor.yml` in both `root_orchestrator/` and `cluster_orchestrator/`; Python remains the default.

Also adds `go_resource_abstractor/client`, a Go HTTP client for this service - the Go-side counterpart to `libraries/resource_abstractor_client` (Python), for future Go consumers of the abstractor.

## Go client library (`go_resource_abstractor/client`)

Thin Go HTTP client covers the same surface as `resource_abstractor_client` (Python). It has consumers for applications, resources (candidates), and jobs, including job-instance sub-routes. Hooks and custom-resources aren’t covered yet.

It’s a Go module nested inside `go_resource_abstractor/` (own `go.mod`, module `github.com/oakestra/oakestra/go_resource_abstractor/client`). Go excludes it from the parent module, so the server build remains unchanged. The client only pulls in the standard library, avoiding dependencies like gin/mongo-driver/testcontainers. This pattern applies to other Go modules in this repo (`scheduler`, `go_node_engine`).

Error handling differs from the Python client. Python returns `None` for 404s, connection failures, or other non-2xx responses, making it unclear if the request failed or if something was not found. This client returns `(value, error)` everywhere: `errors.Is(err, client.ErrNotFound)` for 404s (including `GetByName`, `GetByIP`, and `GetByNameAndNamespace`), `errors.As(err, &apiErr)` for any other non-2xx (`*client.APIError`, status + server message), and plain wrapped errors for transport failures.

Construct a client using `client.New(baseURL, opts…)` or `client.NewFromEnv(opts…)`. The latter reads `RESOURCE_ABSTRACTOR_URL`/`RESOURCE_ABSTRACTOR_PORT`, the same env vars the Python client uses, but errors on missing config instead of silently building `http://None:None`. `WithHTTPClient` and `WithTimeout` compose correctly regardless of order (resolved into a config struct before any `*http.Client` is touched, rather than mutating a client the caller might still hold onto).

Nothing consumes this yet. The scheduler still has its own ad-hoc resource-abstractor HTTP code. `client/README.md` has the adoption recipe (a `replace` directive, plus widening that service’s Docker build context to include `../go_resource_abstractor/client`) for migrating to this client.

## Bug fixes (Go corrects real Python bugs)

- `GET /hooks/<id>` never matched (string-vs-ObjectID comparison) -> always 500. Fixed.
- `PUT /jobs/` fired hooks under `"job"` while every other job route uses `"jobs"`. Normalized to `"jobs"`.
- Job-instance `pre_create`/`pre_update` hook transforms were computed then discarded (re-read of `request.json`). Now applied.
- Concurrent instance-append race could create duplicate instances. Now atomic.
- Custom-resource cascade delete now removes instances before the definition (was reversed; a failed second step could orphan instances).
- `PATCH /hooks/<id>` now rejects invalid event names (marshmallow's field-level validator still runs there in Python despite `validate=False`).
- Numeric fields now persist as BSON integers instead of doubles for plain JSON integers.
- `GET /jobs/<id>?instance_number=n` no longer always 404s (stray top-level filter key removed).

## Remaining behavior differences vs Python - For Reviewers to discuss

| Area | Python | Go | Notes |
|---|---|---|---|
| Missing record (several GET/PATCH/DELETE) | `200` + body `null` | `404` | Absorbed by `resource_abstractor_client` (`raise_for_status()` -> `None` either way). Visible to any direct HTTP caller. |
| Invalid ObjectID on delete | `500` | `400` | More correct, but a status-code change. |
| Single-item resource/hook/custom-resource-definition responses | marshmallow schema strips undeclared fields (e.g. `vram`, `gpu_temp`, `cpu_history`, `memory_history`, `last_modified`) | full DB document returned | List endpoints (`GET /resources/`) already match since Python uses `jsonify` there, no schema. |
| Client-supplied `_id` on create | preserved for candidates/jobs/apps/custom-resource instances | always stripped, server-generates a fresh ObjectID | No current caller supplies `_id` on create. |
| Numeric BSON width | small ints -> BSON `int32` | small ints -> BSON `int64` | Fixes float-vs-int; doesn't match pymongo's width split. Numerically transparent to Mongo queries/JSON. |
| Integers beyond `int64` | pymongo raises (`OverflowError`) | silently widens to `float64` | Not expected in this domain (cpu/memory counts). |
| Numeric/boolean query & body strings | marshmallow coerces `"4"`->`4`, truthy strings->bool | rejected (`422`) | No real client sends stringified scalars. |
| Custom-resource JSON Schema engine | `jsonschema~=4.4` (Python `re`, broad draft support) | `google/jsonschema-go` (RE2, Draft 7/2020-12) | A schema valid under Python can behave differently or fail in Go. Inherent to the rewrite. |
| `GET /jobs?params=x` | `params` replaces the whole Mongo filter -> `500` | `params` ignored, unfiltered list, `200` | Python's branch never did anything useful; not reproduced. |
| Content types / error bodies | `text/html` for string responses, flask-smorest's default `{code,status,errors}` shape | `application/json`, `{message,details}`/`{message}` shapes | Cosmetic; not consumed by field-typed clients. |
| HTTP method surface | 405 on wrong method, HEAD support on GET routes, OPTIONS -> 200, CORS may echo Origin | 404 on wrong method, no HEAD, OPTIONS -> 204 (incl. unknown paths), CORS always `*` | gin/Flask framework defaults; not wired to match. |
| Swagger UI / OpenAPI (`/api/docs`, `/docs/openapi.json`) | served | not served | No internal consumer depends on it. |
| Date serialization | `"2024-01-01 12:00:00+00:00"` (`json.dumps(default=str)`) | RFC3339 (`...T...Z`) | Candidate/job-instance *history* timestamps already match (float / ISO+offset) in both. |

## Testing

- `go_resource_abstractor`: `go build ./...`, `go vet ./...` clean.
- `MONGO_TEST_IMAGE=mongo:7.0 go test ./...` passes (default `mongo:8.0` container doesn't start on this host; documented fallback).
- New regression tests cover each fix above (numeric BSON type, empty-body policy per endpoint, cascade delete order, hook PATCH validation, `instance_number` filter).
- `go_resource_abstractor/client`: `go build ./...`, `go vet ./...`, `go test ./...` clean - no Docker/MongoDB required, the whole suite stubs the abstractor with `net/http/httptest` (success/404/error-body decoding, query-param and request-body shaping, every method's HTTP verb+path, the option-ordering fix, and non-mutation of caller-supplied maps).

## Deployment

- Still opt-in: `export OVERRIDE_FILES=override-go-resource-abstractor.yml` before `StartOakestraRoot.sh` / `StartOakestraCluster.sh` / `StartOakestraFull.sh`.
- Python (`resource-abstractor/`) remains the Compose default.


robertjndw:
## Performance comparison: Python vs Go resource abstractor

Benchmarked both services' official Docker images (built from their actual `Dockerfile`s, unmodified) against isolated `mongo:7.0` instances, using [`hey`](https://github.com/rakyll/hey) for load generation.

**Environment:** macOS (Apple Silicon), Docker/OrbStack, each service + its own Mongo instance on an isolated bridge network (no shared resources between runs).

### Image size & idle memory

| | Python (Flask) | Go | Delta |
|---|---:|---:|---:|
| Docker image size | 183 MB | 18.5 MB | **~10x smaller** |
| Idle RSS (container, post-startup) | 40.1 MiB | 4.1 MiB | **~10x smaller** |

### Throughput & latency (50 concurrent connections)

| Endpoint | Metric | Python | Go | Delta |
|---|---|---:|---:|---:|
| `GET /` (health, no DB) | req/s | 1,575 | 36,424 | **~23x** |
| | avg latency | 30.3 ms | 0.9 ms | **~34x** |
| | p99 latency | 75.0 ms | 2.6 ms | **~29x** |
| `GET /api/v1/applications/` (DB read) | req/s | 1,317 | 19,904 | **~15x** |
| | avg latency | 37.8 ms | 2.4 ms | **~16x** |
| `POST /api/v1/applications/` (DB write) | req/s | 943 | 12,369 | **~13x** |
| | avg latency | 52.5 ms | 3.9 ms | **~13x** |

### Single-connection latency (concurrency=1, isolates per-request overhead)

| | Python | Go | Delta |
|---|---:|---:|---:|
| `GET /` avg latency | 1.1 ms | 0.1 ms | **~11x** |

### Methodology notes

- Both services ran via their real `Dockerfile`s: Python via Flask's built-in dev server (`python resource_abstractor.py`, as actually shipped, no gunicorn/uwsgi in front), Go via `gin` in the distroless image.
- Each got its own freshly-started `mongo:7.0` container, same host, no cross traffic.
- Single measurement run per endpoint, not averaged across multiple trials, so treat these as directionally representative rather than statistically rigorous.
- Some of the concurrency gap is just Flask's dev server being single-threaded by default - that's what the Dockerfile actually runs today, not a rigged test, but it means the concurrent numbers overstate what a real WSGI server (gunicorn + workers) would show against Go. The single-connection numbers are the fairer per-request comparison.

**Takeaway:** the Go port is meaningfully lighter on disk/memory (~10x) and faster per request even without concurrency (~11x), with the gap widening further under concurrent load due to the Python service's single-threaded dev server.

axiphi:
Really cool work, thanks for starting the migration to golang!

I didn't look into individual source files yet, but I have a few remarks on the overall setup of the new resource-abstractor implementation:
- If we're already going for a full refactor, I believe working with an API schema would make a lot of sense. Since we're probably going to stick to RESTful JSON, I guess OpenAPI would be a good fit? There's a codegen library for golang for it: https://github.com/oapi-codegen/oapi-codegen
- I saw you added a golang-based client. I think it makes a lot of sense to have something like that in the future when we migrate other services to it, but right now it would be unused to my understanding. Since it will be present in the Git history for future reference, would you be willing to remove it again @robertjndw? I believe keeping the unused code in the repo to a minimum makes it more manageable.

robertjndw:
> * If we're already going for a full refactor, I believe working with an API schema would make a lot of sense. Since we're probably going to stick to RESTful JSON, I guess OpenAPI would be a good fit? There's a codegen library for golang for it: https://github.com/oapi-codegen/oapi-codegen

You’re right; it might be a good idea. It could make sense to apply a similar process to all future API changes, perhaps even when reviewing/adapting the current Python API.

robertjndw:
> * I saw you added a golang-based client. I think it makes a lot of sense to have something like that in the future when we migrate other services to it, but right now it would be unused to my understanding. Since it will be present in the Git history for future reference, would you be willing to remove it again @robertjndw? I believe keeping the unused code in the repo to a minimum makes it more manageable.

The idea was to have this client already in place when newer parts written in Golang want to directly use the API. This would reduce the burden of rewriting other parts, as they can simply utilize the Client functionality instead of implementing it themselves. I don’t have strong opinions on either approach 😅
Perhaps we should discuss this in the next meeting, as it might be relevant for future refactors as well.

axiphi:
Sounds good, let's discuss this in the next meeting.
I love the OpenAPI-based implementation, thanks a lot!
I will add some additional thoughts as a review.

go_resource_abstractor/api/jobs.go
```diff
@@ -0,0 +1,215 @@
...
+	}
+
+	query := map[string]any{}
+	if n, present, err := queryInt("instance_number", params.InstanceNumber); err != nil {
+		abortInvalidQuery(c, "instance_number", err.Error())
+		return
+	} else if present {
+		query["instance_number"] = n
+	}
+	filter := db.BuildJobFilter(query)
```
axiphi: I think the signature of BuildJobFilter should be changed to accept an instance number directly instead of a map.

go_resource_abstractor/api/resources.go
```diff
@@ -0,0 +1,160 @@
...
+		}
+
+		candidateID, err := s.store.ResolveJobCandidate(ctx, jobID)
+		if abortOnError(c, err) {
+			return
+		}
+		filter["candidate_id"] = candidateID
+	}
+
+	mongoFilter, err := db.BuildCandidateFilter(filter)
```
axiphi, line 43: I think the signature of BuildCandidateFilter should be changed to accept the actual parameter type (or the mapped ones if you want to do the bool coercion right in the request handler).

Right now we put the request params into a map just to unpack them which is unnecessary imo.

axiphi (changes requested):

axiphi:
One more thing to note, which I think we should not address in this PR, but in a future one:
Right now, we use a custom deserialization approach for request bodies with `bindResourceJSONMap`.

I think this is an okay workaround to be compatible with the Python version for now, but comes at the cost of readability and manual type checking.
Therefore I think the better long-term approach here would be to do one of two things:
- Contrary to what Claude thinks (see cfg.yaml), the oapi-codegen library does seem to support additionalProperties correctly, even when generating request body structs (see their [README](https://github.com/oapi-codegen/oapi-codegen#additionalproperties-with-an-object)). So we could use that and manually assemble `AdditionalProperies` into the final MongoDB filter.
- We could abandon the custom filter approach fully. I think there's little benefit to having an abstraction service over MongoDB, if we're just leaking it's implementation right through the API.

When this is merged, we could add another issue for this I guess.

robertjndw:
> One more thing to note, which I think we should not address in this PR, but in a future one: Right now, we use a custom deserialization approach for request bodies with `bindResourceJSONMap`.
> 
> I think this is an okay workaround to be compatible with the Python version for now, but comes at the cost of readability and manual type checking. Therefore I think the better long-term approach here would be to do one of two things:
> 
> * Contrary to what Claude thinks (see cfg.yaml), the oapi-codegen library does seem to support additionalProperties correctly, even when generating request body structs (see their [README](https://github.com/oapi-codegen/oapi-codegen#additionalproperties-with-an-object)). So we could use that and manually assemble `AdditionalProperies` into the final MongoDB filter.
> * We could abandon the custom filter approach fully. I think there's little benefit to having an abstraction service over MongoDB, if we're just leaking it's implementation right through the API.
> 
> When this is merged, we could add another issue for this I guess.

I agree that this rewrite is a bit odd right now. We’re trying to be compatible with the Python library while also improving it. We should discuss our future plans. If this becomes the default resource-abstractor, we could enhance it with better features and ideas, but we also need to update the Python client.

go_resource_abstractor/api/resources.go
robertjndw: I've implemented a fix in the latest commit, would this solution also be fine by you? 😅 
I don't want to have a cross-dependency between the openapi spec and the DB layer, that's why I did not use the parameter type directly

robertjndw:
Superseded by #579.

The branch has been moved from the `robertjndw` fork into `oakestra/oakestra` (same commits, `fc24fd7`) so it can serve as the base of a stacked-PR chain - GitHub requires a PR's base branch to live in the target repo. Since a PR's head cannot be repointed after creation, this one is closed in favour of #579.

Review discussion here is preserved for reference; please continue on #579.

PR #573, opened 2026-08-04
doc: add remark about MongoDB kernel issue

robertjndw:
This pull request updates documentation to clarify and improve troubleshooting steps related to MongoDB 8.0 compatibility issues with specific Linux kernel versions.

**MongoDB 8.0 and Kernel Compatibility:**

* Added a prominent note in `README.md` explaining that MongoDB 8.0 is incompatible with Linux kernel versions 6.19 through 7.0.13, with a link to official documentation.
* Expanded troubleshooting documentation (`SKILLS/troubleshoot-oakestra.md`) to include a new section detailing how MongoDB 8.0 containers may fail to start on these kernel versions, how to check the kernel version, and guidance for users to upgrade the host kernel themselves.
* Updated the troubleshooting table to mention that immediate exit of MongoDB containers without a clear crash trace may be due to this kernel incompatibility.


README.md
```diff
@@ -32,6 +32,8 @@ Worker Node:
 - 100MB RAM
 - ARM64 or AMD64 architecture
 
+*Note: root and cluster orchestrator run MongoDB 8.0, which is incompatible with Linux kernel versions 6.19 through 7.0.13 ([details](https://www.mongodb.com/docs/v8.0/release-notes/8.0/), tracked upstream as [SERVER-125742](https://jira.mongodb.org/browse/SERVER-125742)).*
```
giobart: I think we don't need this any longer right?

giobart (changes requested):

README.md
robertjndw: Not sure, I mean it still does not start but I does not crash anymore 🤔 
CleanShot 2026-09-23 at 09 52 47

giobart: As far as I got by adding the `GLIBC_TUNABLES=glibc.pthread.rseq=1` env we should have that covered. But I don't have a machine with that kernel version to test it.
[Here](https://github.com/appwrite/appwrite/issues/12786) someone else with our same issue.

robertjndw: Ah you're right, then we can remove this remark 👍🏼

robertjndw: I guess we can close this PR then, because the issue seems fixed for now ✅

PR #575, opened 2026-08-23
chore: migrate loggers to log/slog

### LINKED ISSUE #543: Migrate Go loggers to `log/slog`
## Short
Replace the custom `*log.Logger` singleton wrappers in the Go modules with Go's
stdlib structured logger (`log/slog`) for nicer ergonomics and first-class levels/fields. Changes logging API but adds no new dependency.

## Proposal
Both Go modules currently expose accessors that return a `*log.Logger`
(`logger.InfoLogger().Printf(...)`), which is verbose at every call site and offers
no log levels or structured fields as first-class concepts. Levels are faked via
string prefixes (`"INFO-"`) and all context is hand-formatted into the message.

Migrate `go_node_engine/logger/logger.go` and `scheduler/logger/logger.go` to
[`log/slog`](https://pkg.go.dev/log/slog) (Go 1.21+, stdlib), and update all call
sites of `InfoLogger()`, `WarnLogger()`, `ErrorLogger()`, and `DebugLogger()` across
both modules.

Call-site style changes from:
```go
logger.InfoLogger().Printf("deploying %s", name)
```
to:
```go
slog.Info("deploying", "service", name)
```

## Ratio
The current logger is awkward to use (accessor-then-method on every call) and gives
no structured logging. `slog` improves developer ergonomics and log quality while
adding **no** external dependency. It also aligns the two near-duplicate logger packages, which currently
diverge in naming (`infologger`/`infoonce` vs `infoLogger`/`infoOnce`).

## Impact
- NodeEngine (`go_node_engine/logger` + all call sites)
- Cluster/Root scheduler (`scheduler/logger` + all call sites)

No behavioral change to orchestration logic; logging output format and call API change.

## Development time
~1-2 days (mostly mechanical call-site migration + verifying Fatal/debug-toggle behavior).

## Status
Looking for feedback

## Checklist
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

avisoiu:
Closes #543. This pull request replaces the current logging functionality with `log/slog`.
I have opted to keep the abstraction from the logger package in case we would want to make changes to the logging logic in the future without changing the API again. Of course, we can discuss this further.

Therefore, at call sites:

```go

logger.InfoLogger().Printf("deploying %s", name)
```
becomes:

```go

logger.InfoLogger("deploying %s", name)
```

Also introduced a new API for fatal errors, that logs and exits: 

```go
logger.FatalErrorLogger("fatal error")
```

All the module functions now slog in the implementation.

The debug toggle and fatal error behavior are covered by some unit tests.

### DIFF
--- a/go_node_engine/addons/flops/FLOps.go
+++ b/go_node_engine/addons/flops/FLOps.go
@@ -9,7 +9,7 @@ import (
 type FlopsAddon struct{}
 
 func (a FlopsAddon) Startup(configFile []string) {
-	logger.InfoLogger().Printf("Starting FLOps Data Manager")
+	logger.InfoLogger("Starting FLOps Data Manager")
 	HandleFLOpsDataManager()
 }
 
@@ -20,7 +20,7 @@ func HandleFLOpsDataManager() {
 	cmd := exec.Command("docker", "ps", "-a", "--format", "{{.Names}}")
 	output, err := cmd.Output()
 	if err != nil {
-		logger.ErrorLogger().Fatalln("Error:", err)
+		logger.FatalErrorLogger("Error:", err)
 		return
 	}
 
@@ -41,7 +41,7 @@ func HandleFLOpsDataManager() {
 		err := cmd.Run()
 
 		if err != nil {
-			logger.ErrorLogger().Fatalf("Error pulling FLOps Data Manager image: %v\n", err)
+			logger.FatalErrorLogger("Error pulling FLOps Data Manager image: %v\n", err)
 			return
 		}
 
@@ -49,13 +49,13 @@ func HandleFLOpsDataManager() {
 		err = cmd.Run()
 
 		if err != nil {
-			logger.ErrorLogger().Fatalf("Error running container: %v\n", err)
+			logger.FatalErrorLogger("Error running container: %v\n", err)
 			return
 		}
 
 	} else {
-		logger.InfoLogger().Printf("Container %q already exists.", container_name)
+		logger.InfoLogger("Container %q already exists.", container_name)
 	}
 
-	logger.InfoLogger().Printf("FLOps Data Manager container started successfully.")
+	logger.InfoLogger("FLOps Data Manager container started successfully.")
 }
--- a/go_node_engine/virtualization/internal/crosvm/internal/instance/instance.go
+++ b/go_node_engine/virtualization/internal/crosvm/internal/instance/instance.go
@@ -122,12 +122,12 @@ func NewInstance(
 		return nil, err
 	}
 
-	logger.InfoLogger().Printf("setting up network for instance %q...", id)
+	logger.InfoLogger("setting up network for instance %q...", id)
 	netConf, err := setupNetwork(service)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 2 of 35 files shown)


go_node_engine/virtualization/internal/crosvm/internal/instance/instance.go
```diff
@@ -406,9 +406,9 @@ func (i *Instance) waitForExit(cmd *exec.Cmd, startNum uint32) {
 				}
 			}
 
-			logger.ErrorLogger().Print(msgBuilder.String())
+			logger.ErrorLogger(msgBuilder.String())
```
robertjndw, line 409: This passes `msgBuilder.String()` as the format string, not an argument. If the captured crash output contains a `%`, `fmt.Sprintf` will mangle it. Should be `logger.ErrorLogger("%s", msgBuilder.String())`.

go_node_engine/addons/flops/FLOps.go
```diff
@@ -20,7 +20,7 @@ func HandleFLOpsDataManager() {
 	cmd := exec.Command("docker", "ps", "-a", "--format", "{{.Names}}")
 	output, err := cmd.Output()
 	if err != nil {
-		logger.ErrorLogger().Fatalln("Error:", err)
+		logger.FatalErrorLogger("Error:", err)
```
robertjndw, line 23: `"Error:"` has no verb, so `err` never gets formatted in - this will print `Error:%!(EXTRA ...)` instead of the actual error. Needs `logger.FatalErrorLogger("Error: %v", err)`.

go_node_engine/virtualization/internal/crosvm/runtime.go
```diff
@@ -116,7 +116,7 @@ func newRuntime(info virtrt.RuntimeInfo) virtrt.Runtime {
 	_, _ = fmt.Fprintf(&infoMsg, "  > runtime directory: %s\n", runtimeDirPath)
 	_, _ = fmt.Fprintf(&infoMsg, "  > state directory: %s\n", stateDirPath)
 	_, _ = fmt.Fprintf(&infoMsg, "  > cache directory: %s\n", cacheDirPath)
-	logger.InfoLogger().Print(infoMsg.String())
+	logger.InfoLogger(infoMsg.String())
```
robertjndw, line 119: Same as the instance.go spot - `infoMsg.String()` is being used as the format string here too. Should be passed as an argument, e.g. `logger.InfoLogger("%s", infoMsg.String())`.

robertjndw (commented):
Two more things worth a look before merging, not tied to a single line:

- We lose `file:line` on every log line now (old loggers used `log.Lshortfile`). Neither the default handler nor `SetDebugMode`'s `NewTextHandler` sets `AddSource: true`, so with a lot of reused generic messages (`"%v", err` etc.) it'll be harder to tell where a line came from.
- Info used to go to stdout and Warn/Error to stderr; now everything funnels through `slog.Default()` into one stream. Doesn't matter for NodeEngine since systemd already merges both into one file, but it's a real behavior change for the scheduler if anything downstream splits by stream.

PR #578, opened 2026-08-24
540 cluster manager startup

### LINKED ISSUE #540: Improve cluster gRPC startup
## Short
At the cluster orchestrator, to prevent sleeping and hoping the server started, we could try to request the endpoint ourselves on localhost until it becomes available and then register with the system manager. To be fair, if the current solution works, this could be a separate issue.

_Originally posted by @axiphi in https://github.com/oakestra/oakestra/pull/518#discussion_r3419646589_

## Proposal
Test localhost until available before gRPC startup 

## Ratio
More reliable startup

## Impact
Cluster manager

## Development time
1,2 days

## Status
looking for feedback

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

axiphi:
Closes #540


axiphi:
I can't turn this into a PR stack for some reason, so this should be merged after #565 is merged.

robertjndw:
> I can't turn this into a PR stack for some reason, so this should be merged after #565 is merged.

Yeah I noticed this too. PR Stacks don't work with forks yet 🥲

PR #568, opened 2026-07-22
feat: Standardize Python Logging as Structured JSON

Ivoraimar:
> [!NOTE]
> This PR depends on #548 and is based on that PR's branch.
> Please review and merge #548 first.

## Summary

Standardizes Python orchestrator logging around the shared `oakestra_logging` package and schema-v1 JSON contract. Python application, standard-library, and Gunicorn logs now emit one structured stdout record per event with consistent severity, service, context, source, exception, and redaction behavior.

The implementation follows Structlog’s [getting started](https://www.structlog.org/en/stable/getting-started.html), [standard-library integration](https://www.structlog.org/en/stable/standard-library.html), and [bound logger](https://www.structlog.org/en/stable/bound-loggers.html) guidance. JSON is the sole production format so Loki can reliably query records with `| json`.

## Writing logs

Use the shared logger, keep messages static, and pass changing values as structured context:

```python
from oakestra_logging import get_logger

logger = get_logger(__name__)

logger.info(
    "Cluster registration completed",
    event_name="cluster.registration.completed",
    cluster_id=cluster_id,
)
```

When multiple records belong to the same operation, create a bound logger inside that function. The bound fields are automatically included in every record:

```python
operation_logger = logger.bind(
    operation="deployment",
    job_id=job_id,
    cluster_id=cluster_id,
)

operation_logger.info(
    "Deployment started",
    event_name="deployment.started",
)

operation_logger.info(
    "Deployment completed",
    event_name="deployment.completed",
)
```

Keep operation-specific bound loggers local; do not attach request or job data to the module-level logger. Use `logger.exception()` once at the boundary handling a failure, and never log complete payloads, credentials, users, SLAs, MQTT messages, or database records. Full guidance is available in `libraries/oakestra_logging/README.md`.

## Testing

- `pytest libraries/oakestra_logging/tests` — 24 passed
- Ruff and Python compilation
- Compose configuration rendering
- Gunicorn smoke test confirming startup records use schema-v1 JSON

Fixes #566

PR #484, opened 2026-03-15
Return unquoted JSON from Flask API endpoints

HMF2475:
Some API endpoints were returning json.dumps(response) instead of response directly from their Flask handlers, causing the JSON response to be quoted. This PR fixes all affected endpoints to return unquoted JSON, which is the format most developer tools and frameworks expect.

Related issues: [#461 oakestra/oakestra](https://github.com/oakestra/oakestra/issues/461), [#70 oakestra/dashboard](https://github.com/oakestra/dashboard/issues/70), [#31 oakestra/oakestra-cli](https://github.com/oakestra/oakestra-cli/issues/31) [#221 oakestra/oakestra-net](https://github.com/oakestra/oakestra-net/issues/221)

This should be tested in conjunction with the changes in the PR [#223](https://github.com/oakestra/oakestra-net/pull/223) in the Oakestra-net repo and in the PR [#76](https://github.com/oakestra/dashboard/pull/76) in the Dashboard repo.

PR #582, opened 2026-08-26
fix: return the SLA validation error from the SLA POST endpoints

### LINKED ISSUE #171: Return SLA errors
## Short
If there are SLA errors, the exact error should be returned instead of a simple "SLA error".

## Ratio
Facilitates the search for errors

## Impact
Root system manager

## Development time
1 day

## Status

looking for someone to take over

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

hdimer:
Closes #171.

## The problem

Posting an SLA that fails validation never tells you what was wrong. With an SLA whose `microservice_name` is `test.special.char`:

```
POST /api/service/       ->  400 {"code": 400, "status": "Bad Request"}
POST /api/application/   ->  422 {"code": 422, "status": "Unprocessable Entity"}
```

The validator already produces the exact reason, and `create_services_of_app` / `register_app` already return it with the right status code. Three separate things throw it away on the way to the response.

**1. The handler catches its own `abort()`.** In `ServiceControllerPost.post`:

```python
try:
    ...
    result, status = service_management.create_services_of_app(username, data)
    if status != 200:
        abort(status, result)     # raises a werkzeug HTTPException
    return result
except Exception as e:            # ...which is an Exception, so this catches it
    logging.log(logging.ERROR, e)
    abort(400, {"message": "The given SLA was not formatted correctly"})
```

`flask_smorest.abort` raises an `HTTPException`, so the broad handler swallows the abort it just made and replaces every failure with a hardcoded 400. That is why the 422 surfaces as a 400.

**2. The payload is passed positionally.** `flask_smorest.abort` is `abort(http_status_code, exc=None, **kwargs)`, and its error handler only renders `error.data["message"]`, which comes from `**kwargs`. `abort(status, result)` and `abort(code, description=result)` both bind the payload to `exc` and drop it. That is why even the correct 422 from `/api/application/` arrives with no `message` field.

**3. The message is an exception object.** Both service functions return `{"message": e}` where `e` is an `SLAFormatError`. That is not JSON-serializable, so fixing 1 and 2 alone turns the response into a 500. `str(e)` rather than `e.message`, because `SLAFormatError.message` is the generic class attribute this issue is complaining about, and because `v2_validator.validate_json_v2` has an `except ValueError as err: return err` branch that also stores an exception rather than a string.

## The fix

- `services_blueprints.py` - narrow the `try` to the call itself so `abort()` is no longer inside it, and pass `message=`. I moved code rather than adding `except HTTPException: raise`, which would leave the same trap in place for the next non-HTTP raise added inside that block.
- `applications_blueprints.py` - `description=` is not a key flask-smorest renders; use `message=`.
- `service_management.py` / `application_management.py` - `str(e)`, so the validator's text survives serialization.

Every exit from `ServiceControllerPost.post` now carries its message, including the no-body 404 three lines down, which had the same positional-argument bug:

```
POST /api/service/       ->  422 {"code": 422, "message": "'test.special.char' does not match '^[a-zA-Z0-9]{1,30}$'", "status": "Unprocessable Entity"}
POST /api/application/   ->  422 {"code": 422, "message": "'test.special.char' does not match '^[a-zA-Z0-9]{1,30}$'", "status": "Unprocessable Entity"}
POST /api/service/ {}    ->  404 {"code": 404, "message": "/api/deploy request without a yaml file\n", "status": "Not Found"}
```

Only `err.message` from jsonschema reaches the body, so no schema dump leaks out.

## Tests

`tests/blueprints_test.py` is new, and is the first test here that goes through a Flask route. It has to be, for defect 2: that one lives inside flask-smorest's error renderer, so calling the view under a request context would never exercise it. `Api(app)` is the mechanism under test, not fixture ceremony.

The three cases post `tests/service_level_agreements/sla_flawed_5.json`, an existing fixture that `sla_test.py` already pins as raising `SLAFormatError`, and an empty body. I reverted each of the five code changes in isolation to confirm the matching test goes red.

The blueprint imports sit inside the `client` fixture on purpose. pytest collects this file before `services_test.py`, and a module-level import would make it the first importer of `services.service_management`, binding it to the wrong `net_plugin` mock and breaking `services_test.py`'s `assert_called_with`.

Ran on Python 3.10 against the pinned `requirements.txt`: 24 passed. The seven pre-existing `services_test.py` failures remain, all of which need the `resource_abstractor` container that CI starts and I could not run locally. `ruff 0.15.22 check` and `ruff 0.15.22 format --check` pass repo-wide.

## Scope

The sibling `abort()` calls in these two files have the same positional-argument bug, 17 of them. #171 names two endpoints, so this PR fixes those and the one remaining exit inside the same method; fixing an arbitrary subset of the rest would ship untested changes. Happy to follow up on them in a separate PR.

One heads-up: #568 and #484 both touch these lines, for structured logging and JSON quoting respectively. Neither changes this behaviour, but whichever lands second will want a small rebase.

I used AI assistance while writing this; the diagnosis, the tests and the toolchain runs above are ones I checked myself.

### DIFF
--- a/root_orchestrator/system-manager-python/blueprints/applications_blueprints.py
+++ b/root_orchestrator/system-manager-python/blueprints/applications_blueprints.py
@@ -96,7 +96,7 @@ class CreateApplicationController(Resource):
         current_user = get_jwt_identity()
         result, code = register_app(data, current_user)
         if code != 200:
-            abort(code, description=result)
+            abort(code, message=result["message"])
 
         # TODO(ME): Frontend should be able to handle the _id being a string and not an object.
         for i in range(len(result)):
--- a/root_orchestrator/system-manager-python/blueprints/services_blueprints.py
+++ b/root_orchestrator/system-manager-python/blueprints/services_blueprints.py
@@ -112,14 +112,14 @@ class ServiceControllerPost(MethodView):
             try:
                 username = get_jwt_auth_identity()
                 result, status = service_management.create_services_of_app(username, data)
-                if status != 200:
-                    abort(status, result)
-                return result
             except Exception as e:
                 logging.log(logging.ERROR, e)
-                abort(400, {"message": "The given SLA was not formatted correctly"})
+                abort(400, message="The given SLA was not formatted correctly")
+            if status != 200:
+                abort(status, message=result["message"])
+            return result
         logging.log(logging.ERROR, "POST service no data found")
-        abort(404, {"message": "/api/deploy request without a yaml file\n"})
+        abort(404, message="/api/deploy request without a yaml file\n")
 
 
 @servicesblp.route("/<appid>")
--- a/root_orchestrator/system-manager-python/services/application_management.py
+++ b/root_orchestrator/system-manager-python/services/application_management.py
@@ -20,7 +20,7 @@ def register_app(applications, userid):
     try:
         parse_sla_json(applications)
     except SLAFormatError as e:
-        return {"message": e}, 422
+        return {"message": str(e)}, 422
 
     for application in applications["applications"]:
         if app_operations.get_app_by_name_and_namespace(
--- a/root_orchestrator/system-manager-python/services/service_management.py
+++ b/root_orchestrator/system-manager-python/services/service_management.py
@@ -43,7 +43,7 @@ def create_services_of_app(username, data, force=False):
         parse_sla_json(data)
     except SLAFormatError as e:
         logging.log(logging.ERROR, e)
-        return {"message": e}, 422
+        return {"message": str(e)}, 422
 
     app_id = data.get("applications")[0]["applicationID"]
     last_service_id = ""
--- a/root_orchestrator/system-manager-python/tests/blueprints_test.py
+++ b/root_orchestrator/system-manager-python/tests/blueprints_test.py
@@ -1,0 +1,74 @@
+import json
+import pathlib
+from unittest.mock import patch
+
+import pytest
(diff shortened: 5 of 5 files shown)


robertjndw:
Thanks @hdimer, this fixes #171 correctly. 

Non-blocking suggestion: `SLAFormatError` should inherit from `Exception`, and SLA errors could eventually be handled centrally to ensure consistent 422 responses across endpoints. That broader refactor can be a follow-up

@giobart what's your take?

hdimer:
Thanks for the review.

Agreed on both counts, and I'd rather land them separately so this stays a small fix for #171.

On `SLAFormatError(BaseException)`: you're right that it should be `Exception`. I checked that the flip is inert today, both `parse_sla_json` call sites catch `SLAFormatError` by name, so nothing is relying on it slipping past an `except Exception`, and the suite comes out identical either way (24 passed, plus the same 7 pre-existing `services_test.py` failures that need the `resource_abstractor` container). It's a one-word change, so happy to fold it into this PR if you'd rather have it now, or carry it into the follow-up. Your call.

The central 422 handling is the piece I'd keep out. The remaining `abort()` calls in these two files still pass their payload positionally, where flask-smorest drops it, so a shared SLA error handler and that cleanup are really the same job and it wants its own review rather than riding along here.

Happy to open that follow-up once this lands. @giobart, if you'd rather see the shape of it first, I can sketch it on the issue instead.

PR #574, opened 2026-08-12
feat: add Grafana log statistics dashboard

Ivoraimar:
> [!NOTE]
> This PR depends on #564 and is based on that PR's branch.
> Please review and merge #564 first.

## Summary

This PR adds a provisioned Grafana Log Statistics dashboard for Root, Cluster, and 1-DOC, replacing the old mixed page with per-component and per-cluster throughput, warning/error trends, top-N noisy components, and normalized level distribution. It shares filters and time-preserving links with the Logs dashboard.

It also bounds Loki read fan-out and reduces dashboard refresh pressure, preventing concurrent panel queries from exhausting Loki's scheduler and appearing as `No data`; ingestion and stored logs are unchanged.

## Video:

https://github.com/user-attachments/assets/19c20bea-8c91-4e27-8fff-6b48b3350cf6

Fixes #534

PR #583, opened 2026-08-26
feat: add control-plane metrics collection

### LINKED ISSUE #535: Control plane metrics collection pipeline
## Short
Collect real-time host and per-container CPU/memory/disk metrics for Root and Cluster orchestrator components. Foundation for the resources dashboard and resource alerting.

## Proposal
Add a pipeline to collect cluster and root metrics deploying:

- node_exporter for host-level CPU/memory/disk/network on the Root and on each Cluster host.
- cAdvisor for per-container resource usage of the orchestrator services.

Then use Prometheus scraping both exporters at cluster and root.

## Ratio
node_exporter + cAdvisor + Prometheus is the de-facto production standard for container/host monitoring.

## Impact
Cluster and Roto compose files

## Development time
~1 week.

## Status
searching for a solution

## Checklist
 
- [ ] Discussed
- [ ] Documented
- [ ] Implemented
- [ ] Tested

Ivoraimar:
> [!NOTE]
> This PR depends on #574 and is based on that PR's branch.
> Please review and merge #574 first.

## Summary

This PR adds host and container metrics collection to Root, Cluster, and 1-DOC deployments using Prometheus, node_exporter, and cAdvisor. Each deployment stores metrics locally with seven-day/1-GB retention and provisions a non-default Grafana datasource with UID `Prometheus`.

Prometheus and node_exporter remain internal. cAdvisor exposes its diagnostics and `/metrics` endpoint on loopback ports `8081` for Root/1-DOC and `8082` for Cluster. The implementation also covers host-network, IPv6, images-only, and no-observe configurations, including the maintained `run-a-cluster/root-orchestrator.yml` manifest.

**Migration note**: Cluster Prometheus is now named cluster_prometheus and port 10009 is no longer published in normal bridge mode. Grafana accesses it through the private Docker network; the host-network override exposes it only on 127.0.0.1:10009.

Fixes #535

### DIFF
--- a/cluster_orchestrator/docker-compose.yml
+++ b/cluster_orchestrator/docker-compose.yml
@@ -192,18 +192,140 @@ services:
 
 
   prometheus:
-    image: prom/prometheus
-    container_name: prometheus
-    hostname: prometheus
+    image: prom/prometheus:v3.13.2-distroless
+    container_name: cluster_prometheus
+    hostname: cluster_prometheus
     labels: *cluster-log-labels
-    ports:
-      - 10009:9090
+    expose:
+      - "9090"
     volumes:
-      - ./prometheus/:/etc/prometheus/
+      - ./prometheus/prometheus.yml:/etc/prometheus/prometheus.yml:ro
+      - cluster_prometheus_data:/prometheus
     command:
-      - '--config.file=/etc/prometheus/prometheus.yml'
+      - --config.file=/etc/prometheus/prometheus.yml
+      - --storage.tsdb.path=/prometheus
+      - --storage.tsdb.retention.time=7d
+      - --storage.tsdb.retention.size=1GB
     depends_on:
       - cluster_manager
+      - node_exporter
+      - cadvisor
+    extra_hosts:
+      - host.metrics.internal=${CLUSTER_METRICS_GATEWAY:-172.30.251.1}
+    networks:
+      - default
+      - metrics
+    read_only: true
+    tmpfs:
+      - /tmp
+    cap_drop:
+      - ALL
+    security_opt:
+      - no-new-privileges:true
+    restart: unless-stopped
+    logging: *default-logging
+
+  node_exporter:
+    image: quay.io/prometheus/node-exporter:v1.12.1
+    container_name: cluster_node_exporter
+    hostname: cluster_node_exporter
+    labels: *cluster-log-labels
+    network_mode: host
+    command:
+      - --path.procfs=/host/proc
+      - --path.sysfs=/host/sys
+      - --path.rootfs=/host
+      - --collector.filesystem.mount-points-exclude=^/(dev|proc|sys|var/lib/docker/.+|var/lib/containers/storage/.+)($$|/)
+      - --web.listen-address=${CLUSTER_METRICS_GATEWAY:-172.30.251.1}:9100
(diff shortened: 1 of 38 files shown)


root_orchestrator/docker-compose.yml
```diff
@@ -108,6 +108,141 @@ services:
...
+        bind:
+          create_host_path: false
+      - type: bind
+        source: /var/run/docker.sock
+        target: /var/run/docker.sock
+        read_only: true
+        bind:
+          create_host_path: false
+      - type: bind
+        source: ${CONTAINERD_SOCKET:-/run/containerd/containerd.sock}
```
robertjndw, line 219: cAdvisor runs with `--docker_only=true`, so the containerd socket is never read. With `create_host_path: false` this mount is also a hard failure on any host where `/run/containerd/containerd.sock` isn't at that exact path (rootless Docker, Docker Desktop) - compose refuses to create the container and it crash-loops on `restart: unless-stopped`. Think this can just go.

Same mount in the cluster, 1-DOC and root-orchestrator copies.

cluster_orchestrator/docker-compose.yml
```diff
@@ -192,18 +192,140 @@ services:
...
-    image: prom/prometheus
-    container_name: prometheus
-    hostname: prometheus
+    image: prom/prometheus:v3.13.2-distroless
+    container_name: cluster_prometheus
+    hostname: cluster_prometheus
     labels: *cluster-log-labels
-    ports:
-      - 10009:9090
+    expose:
```
robertjndw, line 199: `ports: - 10009:9090` becoming `expose` means cluster Prometheus is no longer reachable off-box. Fine as hardening, but it's breaking for anyone scraping `<cluster-ip>:10009` today (it's documented as a published port), and `container_name` changed to `cluster_prometheus` as well. Worth a note in the description.

robertjndw (commented):
Good to see 1-DOC sharing a single Prometheus/cAdvisor/node_exporter rather than duplicating the stack, and cAdvisor is properly locked down (`read_only`, `cap_drop: ALL`, `no-new-privileges`, trimmed `--enable_metrics`). Two things below.

Ivoraimar:
> Good to see 1-DOC sharing a single Prometheus/cAdvisor/node_exporter rather than duplicating the stack, and cAdvisor is properly locked down (`read_only`, `cap_drop: ALL`, `no-new-privileges`, trimmed `--enable_metrics`). Two things below.

Thanks! I verified the pinned cAdvisor configuration without the containerd socket. Although cAdvisor performs a harmless containerd factory probe during startup, Docker-only CPU and memory collection remains healthy, so I removed the unnecessary mount from all four manifests. I also documented the cluster_prometheus rename and the removal of public port 10009 in normal bridge mode.

Ivoraimar:
Correction to my previous reply: I retested the pinned cAdvisor v0.60.5 with and without the containerd socket. On Docker 29.7.2 using the overlayfs containerd snapshotter, cAdvisor stays alive without the socket, but its Docker factory fails to register and the Compose-labelled container metrics disappear. This matches the [cAdvisor source](https://github.com/google/cadvisor/blob/v0.60.5/container/docker/factory.go#L358-L363). I therefore restored the configurable read-only socket mount in all four manifests and documented the security implication

PR #586, opened 2026-09-07
fix mongo version to 8.0.4

### LINKED ISSUE #572: Document MongoDB incompatibility with Linux kernel 6.19–7.0.13
MongoDB (all packages, including Docker images) is incompatible with Linux kernel versions 6.19 through 7.0.13 due to an incompatibility with the vendored version of TCMalloc. Affected systems will have MongoDB crash/refuse to start.

Since Oakestra runs MongoDB 8.0 as a core dependency in both the root orchestrator (`mongo`, `mongo_net`) and cluster orchestrator (`cluster_mongo`, `cluster_mongo_net`), users deploying on affected kernels will hit this and may not immediately realize it's a kernel issue rather than an Oakestra bug.

From the [MongoDB 8.0 release notes](https://www.mongodb.com/docs/v8.0/release-notes/8.0/):

> **MongoDB Is Incompatible with Linux Kernel 6.19 Through 7.0.13**
>
> Due to an incompatibility between certain Linux kernel releases and the currently vendored version of TCMalloc, running MongoDB with a Linux kernel version from 6.19 through 7.0.13 can cause MongoDB to crash. To prevent this crash, MongoDB detects these kernel versions and stops during startup. This incompatibility affects all MongoDB packages, including those obtained from the MongoDB website, a package manager, or Docker.
>
> Linux kernel version 7.0.14 and later resolves this incompatibility. To run MongoDB on an affected system, upgrade to Linux kernel version 7.0.14 or later.

We should document this, e.g.:
- A note in the main README.
- A check/explanation in `SKILLS/troubleshoot-oakestra.md` (STEP 4, MongoDB diagnostics) so the troubleshooting flow surfaces this quickly.

Fix is to upgrade to Linux kernel 7.0.14+.

giobart:
This pull request updates all MongoDB service definitions across various orchestrator and deployment configuration files to use the more specific `mongo:8.0.4` image instead of the generic `mongo:8.0`. As pointed out by @axiphi this currently is a workaround to avoid the problem described in #572  

Fixes #572 

This has been tested on a device affected by this problem and it worked ;)

**MongoDB image version updates:**

* Updated the `mongo_cluster` and `mongo_clusternet` services in `cluster_orchestrator/docker-compose.yml` to use `mongo:8.0.4`. [[1]](diffhunk://#diff-5154d0b1303f8ab0f078e91c8a9efaf30e058e977d154ed8b2fb4bebc30fdf5cL34-R34) [[2]](diffhunk://#diff-5154d0b1303f8ab0f078e91c8a9efaf30e058e977d154ed8b2fb4bebc30fdf5cL45-R45)
* Updated the `mongo_cluster` service in `cluster_orchestrator/override-images-only.yml` to use `mongo:8.0.4`.
* Updated the `mongo_root` and `mongo_rootnet` services in `root_orchestrator/docker-compose.yml` to use `mongo:8.0.4`. [[1]](diffhunk://#diff-55f28302b0dcb7d5deec53f50f0d6083837cf616577c589ea2a360145b9a889bL51-R51) [[2]](diffhunk://#diff-55f28302b0dcb7d5deec53f50f0d6083837cf616577c589ea2a360145b9a889bL64-R64)
* Updated the `mongo_root` and `mongo_rootnet` services in `root_orchestrator/override-images-only.yml` to use `mongo:8.0.4`.
* Updated the `mongo_root`, `mongo_rootnet`, `mongo_cluster`, and `mongo_clusternet` services in `run-a-cluster/1-DOC.yaml` to use `mongo:8.0.4`. [[1]](diffhunk://#diff-b2a73754621df6bac864b9e557e98b9fdf2a073924f159b11661b5d534b54bf3L60-R60) [[2]](diffhunk://#diff-b2a73754621df6bac864b9e557e98b9fdf2a073924f159b11661b5d534b54bf3L74-R74) [[3]](diffhunk://#diff-b2a73754621df6bac864b9e557e98b9fdf2a073924f159b11661b5d534b54bf3L210-R210) [[4]](diffhunk://#diff-b2a73754621df6bac864b9e557e98b9fdf2a073924f159b11661b5d534b54bf3L224-R224)
* Updated the `mongo_root` and `mongo_rootnet` services in `run-a-cluster/root-orchestrator.yml` to use `mongo:8.0.4`. [[1]](diffhunk://#diff-c9c073a06a6ff0df18815c6503a16b174b9f74b3fdd1fe6a25d6929530525998L40-R40) [[2]](diffhunk://#diff-c9c073a06a6ff0df18815c6503a16b174b9f74b3fdd1fe6a25d6929530525998L53-R53)

### DIFF
--- a/cluster_orchestrator/docker-compose.yml
+++ b/cluster_orchestrator/docker-compose.yml
@@ -31,7 +31,7 @@ services:
       
   # cluster-level mongoDB
   mongo_cluster:
-    image: mongo:8.0
+    image: mongo:8.0.4
     container_name: cluster_mongo
     hostname: cluster_mongo
     ports:
@@ -42,7 +42,7 @@ services:
 
   # cluster-level mongoDB
   mongo_clusternet:
-    image: mongo:8.0
+    image: mongo:8.0.4
     container_name: cluster_mongo_net
     hostname: cluster_mongo_net
     ports:
--- a/cluster_orchestrator/override-images-only.yml
+++ b/cluster_orchestrator/override-images-only.yml
@@ -4,7 +4,7 @@ services:
     pull_policy: always
 
   mongo_cluster:
-    image: mongo:8.0
+    image: mongo:8.0.4
     pull_policy: always
 
   cluster_service_manager:
--- a/root_orchestrator/docker-compose.yml
+++ b/root_orchestrator/docker-compose.yml
@@ -48,7 +48,7 @@ services:
 
   # MongoDB in Root Orchestrator
   mongo_root:
-    image: mongo:8.0
+    image: mongo:8.0.4
     container_name: mongo
     hostname: mongo
     ports:
@@ -61,7 +61,7 @@ services:
 
   # root-level network mongoDB
   mongo_rootnet:
-    image: mongo:8.0
+    image: mongo:8.0.4
     container_name: mongo_net
     hostname: mongo_net
     ports:
--- a/root_orchestrator/override-images-only.yml
+++ b/root_orchestrator/override-images-only.yml
@@ -4,11 +4,11 @@ services:
     pull_policy: always
 
   mongo_root:
-    image: mongo:8.0
+    image: mongo:8.0.4
     pull_policy: always
 
   mongo_rootnet:
-    image: mongo:8.0
+    image: mongo:8.0.4
     pull_policy: always
 
   root_service_manager:
--- a/run-a-cluster/1-DOC.yaml
+++ b/run-a-cluster/1-DOC.yaml
@@ -57,7 +57,7 @@ services:
 
   # MongoDB in Root Orchestrator
   mongo_root:
-    image: mongo:8.0
+    image: mongo:8.0.4
     pull_policy: always
     container_name: mongo
     hostname: mongo
@@ -71,7 +71,7 @@ services:
 
   # cluster-level mongoDB
   mongo_rootnet:
-    image: mongo:8.0
+    image: mongo:8.0.4
     pull_policy: always
     container_name: mongo_net
     hostname: mongo_net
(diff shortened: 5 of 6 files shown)


robertjndw (dismissed):
Maybe we should also add a note to the documentation 🤔

axiphi (dismissed):

robertjndw:
There are now fixes released, MongoDB 8.0.30+ (see [here](https://www.mongodb.com/docs/manual/release-notes/8.0/))
Should we keep the pin, move it or just don't merge it? 👀

giobart:
We might use this v8.0.32 if we add `GLIBC_TUNABLES=glibc.pthread.rseq=1` env. This increases compatibility at the cost of per-thread TCMalloc instead of per-CPU. 
For us jobs statuses have a low amount of operations per second so I would propose addiing this in favor of compatibility with the kernel version discussed above. 

In practice, we update the docker compose files as follows

```
    image: mongo:8.0.32
    environment:
      - GLIBC_TUNABLES=glibc.pthread.rseq=1
```

robertjndw:
> We might use this v8.0.32 if we add `GLIBC_TUNABLES=glibc.pthread.rseq=1` env. This increases compatibility at the cost of per-thread TCMalloc instead of per-CPU. For us jobs statuses have a low amount of operations per second so I would propose addiing this in favor of compatibility with the kernel version discussed above.
> 
> In practice, we update the docker compose files as follows
> 
> ```
>     image: mongo:8.0.32
>     environment:
>       - GLIBC_TUNABLES=glibc.pthread.rseq=1
> ```

Yes, that's fine by me. 👍🏼 
I think in real-world scenarios, it won't have much of an impact.

robertjndw (approved):

robertjndw (approved):

PR #588, opened 2026-09-09
fix: remove FLASK_DEBUG and FLASK_ENV from resource-abstractor Docker…

### LINKED ISSUE #587: FLASK_DEBUG=TRUE in resource-abstractor Dockerfile impacts performance and security
## Short
FLASK_DEBUG=TRUE is enabled in the resource-abstractor Dockerfile, likely by mistake, causing both a security and performance concern.
## Proposal 
The `resource-abstractor/Dockerfile` sets `FLASK_DEBUG=TRUE`, with a comment 
suggesting it's intended for verbose logging:

```dockerfile
# TRUE for verbose logging
ENV FLASK_ENV=development
ENV FLASK_DEBUG=TRUE
```

This appears to be a mix-up between two separate concerns -> `LOG_LEVEL` 
(already present in the Dockerfile) controls logging verbosity, while 
`FLASK_DEBUG` specifically enables Werkzeug's interactive debugger and 
development server behavior.

**Performance impact**: Debug mode adds overhead (auto-reloader, additional 
request handling), which can affect the service's performance.

**Security consideration**: Flask's official documentation notes that debug 
mode allows arbitrary code execution from the browser if an unhandled error 
occurs: https://flask.palletsprojects.com/en/stable/debugging/. While this 
service may not be externally exposed, disabling debug mode is still a good 
security practice.

**Additionally**, the container runs as root by default since there is no 
`USER` instruction in the Dockerfile.

## Solution 
Remove `FLASK_DEBUG=TRUE` and `FLASK_ENV=development` from the Dockerfile, 
and rely solely on `LOG_LEVEL` for controlling logging verbosity. If debug 
mode is genuinely needed for local development, it could be set via a 
separate `docker-compose.override.yml` or passed as a runtime environment 
variable instead of being baked into the image.

Optionally, add a non-root `USER` instruction to avoid running the 
container as root.

## Status
Looking for feedback on the proposed solution before opening a PR.

## Checklist
- [x] Discussed
- [x] Solved
- [x] Tested

omprakash2929:
## What this PR does
Removes `FLASK_ENV=development` and `FLASK_DEBUG=TRUE` from `resource-abstractor/Dockerfile`. These were leftover debug settings baked into the image, which meant the container always ran in Flask's debug mode by default a performance and security concern (debug mode exposes an interactive debugger and stack traces).

Other services (system-manager, cluster-manager) already have `FLASK_DEBUG=FALSE`, so `resource-abstractor` was the odd one out.

Logging is already controlled separately via `LOG_LEVEL`, so no functionality is lost.  only the unnecessary debug flags are removed.

Confirmed with @robertjndw before opening this PR.

Fixes #587

### DIFF
--- a/resource-abstractor/Dockerfile
+++ b/resource-abstractor/Dockerfile
@@ -7,9 +7,7 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . /
 
-# TRUE for verbose logging
-ENV FLASK_ENV=development
-ENV FLASK_DEBUG=TRUE
+
 ENV LOG_LEVEL=WARNING
 
 ENV MONGO_URL=localhost
--- a/resource-abstractor/Dockerfile
+++ b/resource-abstractor/Dockerfile
@@ -7,7 +7,6 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . /
 
-
 ENV LOG_LEVEL=WARNING
 
 ENV MONGO_URL=localhost


resource-abstractor/Dockerfile
```diff
@@ -7,9 +7,7 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . /
 
-# TRUE for verbose logging
-ENV FLASK_ENV=development
-ENV FLASK_DEBUG=TRUE
+
```
robertjndw: Just nitpicking, maybe remove the empty line too Otherwise we have an extra white line 😉

OakAssistant (commented):



All context gathered — the change is a 3-line deletion; I verified runtime behavior, sibling Dockerfiles, compose files, and the entry point.

## Summary

Clean, safe cleanup: `resource-abstractor/Dockerfile` drops `ENV FLASK_ENV=development`, `ENV FLASK_DEBUG=TRUE`, and their misleading comment, leaving `ENV LOG_LEVEL=WARNING` (line 10) as the sole logging knob. This aligns resource-abstractor with system-manager and cluster-manager, and I found no code path that depends on the removed variables. Ready to merge.

## Strengths

- Correct fix for a real inconsistency: `root_orchestrator/system-manager-python/Dockerfile:18` and `cluster_orchestrator/cluster-manager/Dockerfile:24` already ship `FLASK_DEBUG=FALSE`; resource-abstractor was indeed the odd one out.
- No behavior regression: the entry point hard-codes debug off anyway — `resource-abstractor/resource_abstractor.py:79` runs `app.run(host="::", port=RESOURCE_ABSTRACTOR_PORT, debug=False)`, and the container CMD is `python resource_abstractor.py` (Dockerfile:18), not the `flask` CLI, so `FLASK_ENV`/`FLASK_DEBUG` had little effect to begin with. Removing them is pure hygiene with zero runtime risk.
- Logging is genuinely unaffected: `resource_abstractor.py:13-22` reads `LOG_LEVEL` only.
- The stale comment `# TRUE for verbose logging` was removed along with the vars, not left dangling.
- Follow-up commit (`83498a90`) tidied the leftover blank line; the final file reads cleanly.

## Issues (blocking)

None.

## Suggestions (non-blocking)

- `cluster_orchestrator/cluster-manager/Dockerfile:22`, `addons_engine/addons-manager/Dockerfile:11`, `addons_engine/addons-monitor/Dockerfile:11` — these still carry `ENV FLASK_ENV=development` (deprecated in Flask ≥2.2 and meaningless for these CMDs). Worth a follow-up PR to finish the sweep the PR body implies.
- Linked issue #587 also flags the missing `USER` instruction (container runs as root). Out of scope here, but worth tracking so it doesn't get lost when the issue auto-closes via "Fixes #587" — the PR only addresses half the issue.
- `resource-abstractor/start-up.sh:8-9` still exports `FLASK_ENV=development` / `FLASK_DEBUG=False` for local dev — fine to keep (matches the issue's "set debug via runtime env instead" advice), just noting the intentional divergence from the Dockerfile.

## Questions for the author

- The PR body says debug mode was "always on" — strictly, `debug=False` at `resource_abstractor.py:79` already overrode `FLASK_DEBUG`, so the real-world exposure was smaller than stated. Not a problem with the change itself, just confirming you agree this is hygiene/hardening rather than an active vulnerability fix (matters for how #587 gets closed/resolved).
- Should the remaining `FLASK_ENV=development` lines in cluster-manager/addons Dockerfiles land in this PR or a follow-up?

REVIEW_DONE


---
_Tool calls used: 9. To request changes / approvals instead of a comment-only review, a human reviewer should follow up._

giobart (approved):

robertjndw (approved):

